PR #7, opened 2023-01-17
[Ng Jia Yi] iP

gremmyz:
## DukePro
>"Your mind is for having ideas, not holding them." – David Allen ([source](https://dansilvestre.com/productivity-quotes))

- text-based
- easy to learn

All you need to do is,
1. download it from [here](https://github.com/gremmyz/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---
```java
public class Main {
    }
}
```


src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task {
+
+    protected String from;
+    protected String to;
```
evanpy: "from" and "to" might be ambiguous, well at least from my pov. Instead of time, readers could think of it in terms of place.
Other than that, nice code 👍🏻 good to go.

evanpy (commented):
LGTM! 😄

src/main/java/Event.java
gremmyz: Thanks for the feedback! You're right, will adjust as necessary 👍🏼

lists/taskList.txt
```diff
@@ -0,0 +1,3 @@
+T | 0 | yeet
+T | 0 | eat some bread
+E | 0 | pet some cats | 2023-12-02
```
junianob: Maybe you might want to keep your taskList private?
You can consider adding lists/taskList.txt to your .gitignore file!

src/main/java/seedu/duke/Event.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Runs the Event creation
+     * Checks the number of tasks in the list and prints it out.
+     *
+     * @param taskList    the TaskList to add the Event object into.
+     * @param description String input for description of an Event object, with task name and event date.
+     *
+     * @return String for Event creation
+     */
+    public static String runEvent(TaskList taskList, String description) {
```
junianob, line 54: You used description for the variable name here and desc somewhere else, would it be preferable to stick to a consistent naming scheme?

src/main/java/seedu/duke/Todo.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Creates a new Todo object based on the user input in the description.
+     * The description is the task name.
+     *
+     * @param taskList the TaskList to add the Todo object into.
+     * @param desc     String input for description of a Todo object.
+     *
+     * @return String for Todo creation.
+     */
+    public static String createTodo(TaskList taskList, String desc) {
```
junianob: You used desc for the variable name here and description somewhere else, would it be preferable to stick to a consistent naming scheme?

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+            // Set command
+            String command = description[0];
+            Commands currCommand = Commands.valueOf(command.toUpperCase());
+
+        switch (currCommand) {
+            // Command for bye
+            case BYE:
+                response = ui.displayExit();
+                break;
+                // Command for list
```
junianob: not inline! Maybe wrote the comment on the wrong line?

src/main/java/seedu/duke/Ui.java
```diff
@@ -0,0 +1,159 @@
...
+        if (taskList.size() == 0) {
+            return "You have no tasks. Hooray!";
+        } else {
+            int index = 1;
+            String list = "";
+            while (index < (taskList.size() + 1)) {
+                list = list + "\n   " + index + ". " + taskList.get(index - 1).toString();
+                index++;
+            }
+            String hasTasks = "Here are your tasks:";
```
junianob: Would it be cleaner to return in a single statement without the hasTasks String?

src/main/java/seedu/duke/Ui.java
```diff
@@ -0,0 +1,159 @@
...
+        if (taskList.size() == 0) {
+            return "There are no such tasks.";
+        } else {
+            int index = 1;
+            String list = "";
+            while (index < (taskList.size() + 1)) {
+                list = list + "\n   " + index + ". " + taskList.get(index - 1).toString();
+                index++;
+            }
+            String hasTasks = "Here are the tasks with the keyword: ";
```
junianob: Would it be cleaner to return in a single statement without the hasTasks String?

junianob (commented):
LGTM! Coding standard is followed and class, method and fields are named appropriately. Minor nitpicks/suggestions that can be trivially handled. :)

src/main/java/seedu/duke/Event.java
gremmyz: Thanks for the feedback! I agree with you, will make the necessary amendments.

src/main/java/seedu/duke/Parser.java
gremmyz: Fixed it!

src/main/java/seedu/duke/Ui.java
gremmyz: Amended!

src/main/java/seedu/duke/Ui.java
gremmyz: I agree with you, my original intention was so that the return statement would be in a single line for me to read it easily 😄

PR #19, opened 2023-01-19
[zenithyap] iP

zenithyap:
# Pix User Guide :poop: 
> “Your mind is for having ideas, not holding them.” – David Allen  

  **Pix helps to remember stuff for you so that you do not have to! It is a Command-Line-Interface based task manager that is ~~easy~~ super easy to learn and quick to use.**

- [Installation](#installation)
- [Features](#features)
- [Commands](#commands)

## Installation
1. Ensure that you have java `11` installed.
2. Download the [ip.jar](https://github.com/zenithyap/ip/releases/download/Level-9/ip.jar) file.
3. Open a terminal and navigate to the folder the file is downloaded in.
4. Run the command `java -jar ip.jar`
5. You should see the following:
 ```
 Hello, this is
 ______    __  ___   ___ 
 |   _  \  |  | \  \ /  / 
 |  |_)  | |  |  \  V  /  
 |   ___/  |  |   >   <   
 |  |      |  |  /  .  \  
 | _|      |__| /__/ \__\

 --------------------------------------------------------------
  How can I assist you?
 --------------------------------------------------------------
 ```

### Manages Your Tasks
- [X] Able to add/delete tasks
- [X] Supports 3 tasks, todo, deadlines, and events
- [X] Mark/Unmark tasks 
- [X] Find tasks by keyword

### `todo`  
Adds a todo task.  
  
Format: `todo <Description>`

### `dline`  
Adds a deadline task. By to be specified as yyyy-MM-dd HH:mm.
  
Format: `dline <Description> / <By>`

### `event`  
Adds an Event task. From and To to be specified as yyyy-MM-dd HH:mm.
  
Format: `event <Description> / <From> / <To>`

### `ls`  
Lists all the tasks.
  
Format: `ls`

### `lsd`  
Lists deadlines that are due by the date and events that area ongoing during the date. Date format as yyyy-MM-dd 
  
Format: `lsd <Date>`

### `mk`  
Marks a task as done.
  
Format: `mk <Index>`

### `unmk`  
Unmarks a task.
  
Format: `unmk <Index>`

### `rm`  
Removes a task from the list.
  
Format: `rm <Index>`

### `find`  
Find tasks that contains a keyword.
  
Format: `find <Keyword>`

### `bye`  
Exits pix.
  
Format: `bye`
  
# Pix's main code:
```Java
public class Pix {
    /**
     * Main function which runs Pix's logic.
     *
     * @param args Arguments provided.
     */
        MyData data = new MyData();
        Ui ui = new Ui();
        try {
            data.loadData();
        } catch (IOException e) {
            System.out.println(e.getMessage());
        }
        Parser parser = new Parser(data);
        Ui.display();

        Scanner sc = new Scanner(System.in);

        while (sc.hasNextLine()) {
            String command = sc.nextLine();
            try {
                Command parsed = parser.parse(command);
                parsed.execute(data, ui);
            } catch (PixException e) {
                System.out.println(e.getMessage());
            }
            if (command.equals("bye")) {
                break;
            }
        }
        sc.close();
    }
}
```


src/main/java/pix/commands/AddDeadline.java
```diff
@@ -0,0 +1,36 @@
+package pix.commands;
+
+import pix.data.MyData;
+import pix.tasks.Deadline;
+import pix.ui.Ui;
+
+/**
+ * AddDeadline class to add deadline to list of tasks.
+ */
+public class AddDeadline extends Command {
```
jerrrren: Class name should be noun so wouldn't DeadlineCommand would be better?

src/main/java/pix/commands/Bye.java
```diff
@@ -0,0 +1,13 @@
+package pix.commands;
+
+import pix.data.MyData;
+import pix.ui.Ui;
+
+/**
+ * Bye class which exits pix.
+ */
+public class Bye extends Command {
```
jerrrren: Class name should be noun so wouldn't ByeCommand would be better?

src/main/java/pix/data/MyData.java
```diff
@@ -0,0 +1,136 @@
...
+import pix.tasks.Event;
+import pix.tasks.Task;
+import pix.tasks.ToDo;
+
+/**
+ * MyDate class to keep track of the list of tasks.
+ */
+public class MyData {
+    /** Arraylist to keep track of all tasks. */
+    private final ArrayList<Task> data = new ArrayList<Task>();
```
jerrrren: Is it more appropriate to name this variable as tasks since it is an arraylist consisting of tasks which is the plural of task?

src/main/java/pix/data/MyData.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public int len() {
+        return this.data.size();
+    }
+
+    /**
+     * Saves the list of tasks to a text file on local device.
+     */
+    public void saveToFile() {
+        String home = System.getProperty("user.dir");
```
jerrrren: Variable name can be more specific instead of something generic like home, possibly name it like homeDirectory?

src/main/java/pix/tasks/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+/**
+ * Deadline task containing the description and due date.
+ */
+public class Deadline extends Task {
+    /** Date and time of deadline. */
+    protected LocalDateTime by;
```
jerrrren: If possible try to replace this variable with another noun?

jerrrren (commented):
In general, good job in ensuring that formatting follows the java standard. However all the command subclasses are not named in a noun fashion so maybe it would be better for you to use an alternative class name.

src/main/java/pix/commands/AddEvent.java
```diff
@@ -0,0 +1,37 @@
+package pix.commands;
+
+import pix.data.MyData;
+import pix.tasks.Event;
+import pix.ui.Ui;
+
+/**
+ * AddEvent class to add event to list of tasks.
+ */
+public class AddEvent extends Command {
```
hhchinh2002: Would naming classes representing different types of tasks and event as "AddSomething" lower the class dealing scopes? E.g In the future, the class may need to encapsulate some other information about the task, the class name may sound confused in that case since it suppose to use as adding event/task only.

src/main/java/pix/parser/Parser.java
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Parses deadline command.
+     *
+     * @param slashed Array of description and due date.
+     * @return AddDeadline object.
+     * @throws PixException If no description, due date or invalid format of date given.
+     */
+    public Command parseDeadline(String[] slashed) throws PixException {
+        if (slashed.length != 2 || removeCommand(slashed[0]).isEmpty() || removeCommand(slashed[1]).isEmpty()) {
```
hhchinh2002: Would it be more readable if these conditions is  initialized as seperated boolean variables instead invoking methods inside if statement?

src/main/java/pix/parser/Parser.java
```diff
@@ -0,0 +1,260 @@
...
+            throw new PixException("Invalid format.\n\n"
+                    + "e.g. deadline 'description' / 'yyyy-MM-dd HH-mm'");
+        }
+        try {
+            return new AddDeadline(removeCommand(slashed[0]), removeCommand(slashed[1]));
```
hhchinh2002: Same as above comment, since removeCommand(Slashed[0]) and removeCommand(Slashed[1]) is referred to twice, would it be neater if we initialize these 2 objects with a more readable name and calling to it when needed?

hhchinh2002 (commented):
Overall, the codes are great, just some small suggestions to look more readable, I learned a lot from doing reviews on your code,

PR #32, opened 2023-01-20
[immanuelhume] iP

immanuelhume:
# Chungus

> "It's Morbin Time." - Dr. Michael Morbius

*Chungus* (not to be confused with [Big Chungus](http://artsatmichigan.umich.edu/ink/2019/01/11/who-is-big-chungus/)) is a simple task tracking assistant. Chungus is

* Text-based
* Easy to learn
* Written in **Java 11**, one of the languages of all time :weary: 

## Basic Usage

1. Download the latest release [here](https://github.com/immanuelhume/ip/releases)
2. Ensure Java 11 is installed
3. Run *Chungus* with `java -jar chungus.jar`

## Reading the Code

The entry point can be found in `Chungus.java`.

```java
class Chungus {
    ...
        new Chungus(System.in, System.out, DEFAULT_DB_PATH).spin();
    }
    ...
}
```

Each Chungus instance expects an input and output stream, as well as the path to a file on disk to store its data. By default, it uses `./chungus.db`, relative to the current working directory.

## Feature Roadmap

- [x] Manage different kinds of tasks - todo, deadlines, and events
- [ ] Cure depression (coming soon)


Lanreath (commented):
Appreciate the commitment to chonciness 👏 👏

src/main/java/chungus/Task.java
```diff
@@ -0,0 +1,405 @@
...
+
+    public String[] getTags() {
+        return tags.stream().toArray(String[]::new);
+    }
+}
+
+/**
+ * A todo.
+ */
+class Todo extends Task {
```
ginloy, line 158: Perhaps put the different tasks in separate files? And then put them all in one separate package.

src/main/java/chungus/util/Pair.java
```diff
@@ -0,0 +1,51 @@
+package chungus.util;
+
+/**
+ * A general pair class.
+ */
+public class Pair<S, T> {
+    private S first;
+    private T second;
```
ginloy: Not essential but consider making these private attributes final since there are no setters.

src/main/java/chungus/Parser.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Parses a command, and selects a suitable handler. This is a pure function.
+     * The handler returned is not safe to be ran multiple times.
+     * 
+     * @param cmd The raw command to parse.
+     * @return An appropriate handler for the command.
+     * @throws ChungusException When the expected command format is wrong.
+     */
+    public static Handler parse(String cmd) {
```
ginloy, line 28: Very long method. Perhaps abstract away the parsing so that it is easier to maintain and add new cases. But nice use of switch cases!

src/main/java/chungus/Chungus.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * A task management app.
+ */
+public class Chungus {
+    private NonBlockingUi ui;
+    private TaskList tasks;
+    private Storage db;
+    private AtomicBoolean isRunning;
+
+    public static final String DEFAULT_DB_PATH = System.getProperty("user.dir") + "/chungus.db";
```
ginloy, line 18: Perhaps the static attributes should be above the private final attributes according to coding standards

src/main/java/chungus/ChungusException.java
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * An exception to represent when the serialization of a task is invalid.
+ */
+class TaskMarshalException extends ChungusException {
+    public TaskMarshalException(String s) {
+        super(String.format("Bad marshal format: \"%s\"", s));
+    }
+}
```
ginloy, line 32: Same as tasks, maybe put these into separate files and put all the exceptions into one package.

src/main/java/chungus/Task.java
```diff
@@ -0,0 +1,405 @@
...
+/**
+ * Represents a task.
+ */
+abstract class Task {
+    private String desc;
+    private boolean isDone;
+
+    protected HashSet<String> tags;
+
+    protected static final DateTimeFormatter DATETIME_FMT = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
```
ginloy, line 18: Same here, perhaps the static attributes should be above the private final attributes according to coding standards

ginloy (commented):
Overall code looks gr8 with very detailed documentation. Just some minor formatting and organization issues. Consider separating the io, exceptions, tasks into different packages.

src/main/java/chungus/Chungus.java
immanuelhume: In which page did you find this? I wasn't able to find it at https://se-education.org/guides/conventions/java/intermediate.html

src/main/java/chungus/util/Pair.java
immanuelhume: Good idea

src/main/java/chungus/Parser.java
immanuelhume: I think this is fine for now as the length comes from the number of cases. Within each case we only extract some values and pass it to the handler so it's not actually doing much. Thanks for the suggestion though!

src/main/java/chungus/Chungus.java
ginloy: Tbh idk someone suggested that to me too.

immanuelhume: lol

PR #36, opened 2023-01-20
[KongHeyi] iP

heeeyi:
## Duke Plus

> Software and cathedrals are much the same; first we build them, then we pray. – by Anonymous

Duke Plus is a task-management app, with the attentive design of saving your memory losses out of long-term intensive coding. If you are also worried about yourself forgetting things, click [here](https://xkcd.com/1093/) to find some consolations.

Aside from the fact that you yourself need to go through stressful coding to produce it, as a user, you will feel it 

* *Easy* to use 😄 
* Friendly to **typewriting-lovers**
* ~~Not friendly to people too lazy to type the task~~

What you need to do to use it is,
1. Download it (coming soon)
2. Double-click to run
3. Type your tasks to add
4. Ask it to remind you when tasks are coming up

Features:

- [x] Adding, deleting, marking tasks
- [x] Find task based on time and keyword
- [ ] Reminders (coming soon)

I am not sure you want to see coding anymore if it has tortured you so much that you need the app to remind you things, but this is the `main` method for the app:

```
public class Main {
    }
}
```


src/main/java/dukes/task/Task.java
```diff
@@ -0,0 +1,95 @@
...
+        this.tag = "";
+    }
+
+    public void setDone() {
+        this.isDone = true;
+    }
+
+    public void setUnDone() {
+        this.isDone = false;
+    }
```
PeanutButters93, line 43: Perhaps consider combining these setter methods into 1?
public void setDone(boolean isDone)

src/main/java/dukes/task/Task.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public boolean hasDone() {
+        return this.isDone;
+    }
```
PeanutButters93, line 47: Function name fits convention but sounds grammatically weird. Might I suggest getDoneStatus()

src/main/java/dukes/util/UI.java
```diff
@@ -0,0 +1,235 @@
...
+
+import java.util.Scanner;
+
+/**
+ * The util class for reading inputs and providing feedbacks to user.
+ */
+public class UI {
+    // mainly use static methods
+    // need to handle the exceptions raised by Parser
+    public static final String WORD_DIVISHION_LINE = "________________________________________";
```
PeanutButters93, line 13: I think this might be a typo, might want to spell check division for clarity. But otherwise, I like the use of the constants. Very clean and neat looking.

src/main/java/dukes/command/ListCommand.java
```diff
@@ -0,0 +1,149 @@
...
+            for (int i = 0; i < taskList.size(); i++) {
+                Task theTask = taskList.get(i);
+                if (theTask.getTag().equals("D") &&
+                        theTask.getDeadLine().equals(date)) {
+                    counter += 1;
+                    sb.append(counter).append(". ");
+                    sb.append(theTask.toString()).append("\n");
+                } else if (theTask.getTag().equals("E") &&
+                        (theTask.getStart().isBefore(date) || theTask.getStart().equals(date)) &&
+                        (theTask.getEnd().isAfter(date) || theTask.getEnd().equals(date))) {
```
PeanutButters93: This function has few comments and almost no line breaks. I find it a bit difficult to understand - especially this long boolean expression. Perhaps consider adding some line breaks and commenting this part?

PeanutButters93 (commented):
Looks v good and clean!
LGTM

src/main/java/dukes/engine/Duke.java
```diff
@@ -0,0 +1,106 @@
...
+            ui.showLoadingError(e.getMessage());
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Runner of Duke class. Call classes to interpret command and execute.
+     * Catch all the runtime DukeExceptions and provide feedback.
+     */
+    public void run() {
```
FredericChow00: I think since we are using the GUI and FXML for the handling of inputs, there is no longer a need for the use of Scanners since the front end of the GUI will call the handleUserInput() method in MainWindow.java. Perhaps this could be removed?

src/main/java/dukes/command/Command.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Abstract method that passes the wrapped information for execution to TaskList, UI and Storage.
+     *
+     * @param tasks contains the task list.
+     * @param ui the UI in charge of user interactions.
+     * @param storage handles the loading and saving of files.
+     * @throws DukeException if unexpected runtime issue occurs.
+     */
+    public abstract void execute(TaskList tasks, UI ui, Storage storage) throws DukeException;
```
FredericChow00: Seems like `execute(...)` and `runCommand(...)` have very similar purposes. 

Imo you can just omit `execute(...)` since this is used in the `run()` method in Duke.java with Scanners etc but there isn't a need for that anymore with the GUI

src/main/java/dukes/engine/Duke.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /**
+     * Main driver of the Duke engine.
+     *
+     * @param args keyboard arguments
+     */
+    public static void main(String[] args) {
+        new Duke("data/tasks.txt").run();
```
FredericChow00: Perhaps you can replace this file path with the static class variable `DEFAULT_PATH`? Since it is already defined at the start

FredericChow00 (commented):
Good job! In general there is quite a substantial amount of OOP and abstraction such as having different classes for each type of Command. 

There are some parts of the codebase that could be ommitted though from what I am seeing, such as the use of Scanners to handle inputs. Other than that, everything else looks fine to me!

PR #46, opened 2023-01-20
[Li Xintong]iP

Emrysil:
# SebastianMax
> Hello World!

SebastianMax frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is:
1. download it from [here](https://github.com/Emrysil/ip/releases/tag/A-Jar)
2. double click it
3. add your tasks
4. let it manage your tasks for you 💯 
   
And it is **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you are a programmer, you can use it to practice Java too! Here is the `Main` method
```Java
}
```


src/main/java/controllers/DialogBox.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        dialog.setText(text);
+        displayPicture.setImage(img);
+        this.style();
+    }
+
+    /**
+     * Styles the dialog box and other components it contains
+     */
+    private void style() {
```
zxisatree: Perhaps this method could be named applyStyle instead? style could be a noun as well, and the meaning of this method might be ambiguous.

src/main/java/controllers/Main.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+import sebastian.Sebastian;
+
+
```
zxisatree: Is more than one line of whitespace necessary here?

src/main/java/sebastian/command/ExitCommand.java
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * To exit the current session
+     * @param taskList taskList instance created at the start of the session
+     * @param ui ui instance created at the start of the session
+     * @param storage storage instance created at the start of the session
+     * @return a string representing the result of task execution
+     */
+    @Override
+    public String execute(TaskList taskList, Ui ui, Storage storage) {
+        String res = "Bye. It's my pleasure to serve you";
```
zxisatree: The variable `res` seems unneccessary, as it is only used once. Might be simpler to just return the string directly.

src/main/java/sebastian/main/Parser.java
```diff
@@ -0,0 +1,55 @@
...
+import sebastian.command.UnmarkCommand;
+import sebastian.exceptions.IllegalInputException;
+
+/**
+ * A class to interpret user input
+ */
+public class Parser {
+
+    /**
+     * To interpret user command and dispatch the job to different command handlers for execution
```
zxisatree, line 23: Perhaps "Interprets user commands..." instead of "To interpret user command..." is more descriptive and follows the imperative tone that the coding standard suggests.

src/main/java/sebastian/task/Deadline.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param taskDescription description of the deadline
+     * @param endTime the time the deadline is due
+     */
+    public Deadline(int isCompleted, String taskDescription, EndTime endTime) {
+        super(taskDescription, isCompleted);
+        this.endTime = endTime;
+    }
+
+    /**
+     * Checked if the deadline is on the same day as the provided date
```
zxisatree, line 31: "Checks if..." instead of "Checked if..." could be used to follow the coding standard's tone.

src/main/java/sebastian/time/DatePattern.java
```diff
@@ -0,0 +1,21 @@
...
+ * Enum class specifying different date time patterns for uses under different conditions
+ */
+public enum DatePattern {
+    PRESENTATION_FORMAT("dd MMM yyyy HH:mm"),
+    USER_INPUT_FORMAT("yyyy-MM-dd HHmm"),
+    TASK_ON_DATE_FORMAT("yyyy-MM-dd");
+
+    private final String format;
+
+    DatePattern(String format) {
```
zxisatree, line 13: Missing access modifier? Not sure if the constructor is intended to be default.

zxisatree (commented):
I like the separation of code into the various packages, makes it very easy to navigate and find relevant code. Overall very neat and comprehensive code.

src/main/java/sebastian/main/Storage.java
```diff
@@ -0,0 +1,87 @@
...
+        this.file = new File(filePath);
+    }
+
+    /**
+     * Load data from hard disk and convert the String representation into actual tasks.
+     * @return a main.TaskList containing all the tasks.
+     */
+    public TaskList formTaskListFromData() throws CannotLoadDataException {
+        TaskList taskList = new TaskList();
+        File file = new File("src/main/resources/SebastianData.txt");
```
Jxleejiaxin: Perhaps it might be better to extract out the file path for your Storage class methods? Your Storage class constructor receives a String _filePath_ as argument and creates a _file_ field but it is immediately modified in your methods. It is also more convenient if you decide to change your file path in the future.

Jxleejiaxin (commented):
I like how you have extracted so many different kinds of exceptions. I think more system outputs could be extracted to the UI class. Keep up the good work!

PR #51, opened 2023-01-21
[FredericChow00] iP

FredericChow00:
# DukeBaby

> Due tomorrow do tomorrow - CHS 4-1 2016

DukeBaby is the personal assistant that everybody wants and needs to stop them from procrastinating

* Easy to learn
* Easy to use
* ~~Easy~~ _VERY EASY_ way to cure procrastination

All you need to do is:
1. Download from [here](https://github.com/FredericChow00/ip/releases).
1. Double-click it.
1. Add your tasks
1. ~~Start~~ Stop your procrastinating journey! :tada:

Best of all, it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Managing deadlines (coming soon!)
- [ ] Reminders (coming soon!)

If you are a Java programmer, here is the `main` method:

``` ruby
public class Main {
    }
}
```


src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,47 @@
...
+    private TaskList tasks;
+
+    private static final String[] LIST_OF_COMMANDS =
+            new String[]{"list", "bye", "todo", "mark", "unmark", "event", "deadline", "delete", "find"};
+
+    private static final String SAVE_FILE_NAME = "duke.txt";
+
+    enum Commands {
+        list, bye, todo, mark, unmark, event, deadline, delete, find
+    }
```
WindyCall: Will this be better if you use a separate Enumeration class instead of putting it inside your main Duke class?

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,262 @@
...
+                Task deletedTask = taskList.get(index);
+                updatedList = taskList.deleteTask(index);
+                storage.writeFile(updatedList);
+                output = ui.sayDeletedTask(deletedTask, updatedList);
+                break;
+            case find:
+                String keywords = getKeywords(inputStrings);
+                TaskList matchingTasks = taskList.findTask(keywords);
+                output = ui.sayMatchingTasks(matchingTasks);
+                break;
```
WindyCall, line 399: Very good follow of switch/case coding standard!!

src/main/java/seedu/duke/TaskList.java
```diff
@@ -0,0 +1,143 @@
+package seedu.duke;
+
+import seedu.duke.tasks.*;
```
WindyCall: Should explicitly importing all classes needed better as required by given Java coding standard?

WindyCall (commented):
Well done!! I think your variable namings are generally very well considered and self-explanatory. What impressed me most is how your format corresponds to the requirement and your good overall OOP construction. Nice to see so excellent work!!!

src/main/java/seedu/duke/TaskList.java
FredericChow00: Yup will change this! I used * because I thought that if there's more than 3 imports from the same package we can just replace with * but i guess not hahaha

src/main/java/seedu/duke/Duke.java
FredericChow00: Hmm yeah that's true although I'm not sure if my Enum class will have any methods though. Thanks for the advice!

src/main/java/seedu/duke/Main.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+import seedu.duke.gui.MainWindow;
+
```
jytheawesome, line 11: Do you need to separate imports of different packages by a blank line here? Correct me if I'm wrong

src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,47 @@
+package seedu.duke;
+
+public class Duke {
+    private final Storage storage;
+    private final Ui ui;
+    private final Parser parser;
+    private TaskList tasks;
```
jytheawesome, line 7: good use of access modifiers!

jytheawesome (commented):
I liked your naming conventions! They were clear and easily understandable. Good job!

PR #52, opened 2023-01-21
[Tang Bo Kuan] iP

bokuanT:
# Duke bot
> A duck a day means three ducks on wednesday
- _text-based_
- easy to use
- **never** forget your deadlines

All you need to do is:

1. download it from ~~there~~ [here](https://github.com/bokuanT/ip/releases)
2. run the jar file
3. add tasks
4. **profit**

fenced code block
```java {.line-numbers highlight-lines="1[:]"}
public class Main {
        System.out.println("hello world");
    }
}
```

- [x] todo, deadlines and events
- [ ] contacts
- [ ] Add delight to the experience when all tasks are complete :tada:

inline code `public static void main(String[] args)`


data/tasks.txt
```diff
@@ -0,0 +1,2 @@
+E@false@book fair@11/01/2000 1000@12/01/2000 1200
+D@false@task a@11/11/2000 1900
```
jinnieshin: Would you add this file to .gitignore so that the user data would not be revealed?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,77 @@
...
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Starts Duke's backend processes.
+     */
+    public void run() {
+        ui.showInitMessage();
+        ui.showGreeting();
```
jinnieshin, line 59: Is there a reason why you separated the InitMessage and Greeting into separate functions?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + Task.getDateTimeString(this.by) + ")";
+    }
+
+    @Override
+    public String getFileFormatString() {
+        //to be split using "|"
+        return "D" + "@" + this.isDone + "@" + this.description + "@" + this.byString;
+    }
```
jinnieshin: I am not sure about what do you mean by 'to be split using "|"' as there is no "|" in your code. Could you please explain? I noticed the same issue in Event and ToDo classes too.

text-ui-test/ACTUAL2.TXT
```diff
@@ -0,0 +1,21 @@
...
+	____________________________________________________________
+	1. [T][X] read book
+	2. [D][ ] return book (by: 06 Jun 2023 1800)
+	3. [T][X] join sports club
+	4. [T][ ] borrow book
+	5. [E][ ] project meeting (from: 22 Jan 2023 0800 to: 22 Jan 2023 2200)
+	____________________________________________________________
+	____________________________________________________________
+	Bye. Hope to see you again soon!
+	____________________________________________________________
```
jinnieshin: I like that you have been conducting TextUiTest continuously👍 I should also do this.

jinnieshin (commented):
Looks good. I like how you took care of small details!

data/tasks.txt
bokuanT: Thank you for catching that! It would indeed be bad if the user’s data is leaked

src/main/java/duke/task/Deadline.java
bokuanT: This is a mistake on my part for not updating the comments

src/main/java/duke/Duke.java
bokuanT: I didn’t have a particular reason, but after the addition of GUI, I may not want the init message to be said by duke anymore as it isn’t really ‘speakable’

PR #54, opened 2023-01-21
[daitenshionyan] iP

daitenshionyan:
Aqua frees your mind of having to remember things you need to do. She is,
* Text-based
* Easy to learn
* ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1. Download it from [here](https://github.com/daitenshionyan/ip/releases).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉

And she is **CUTE**!

Features:
- [x] Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)
________________________________________
If you are a Java programmer, you can use it to practice Java too. Here is the `main` method:
```java
public class Aqua {
    }
}
```


src/main/java/aqua/MainApp.java
```diff
@@ -0,0 +1,20 @@
...
+import javafx.application.Application;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+
+
+/** Main application class. */
+public class MainApp extends Application {
+    @Override
+    public void start(Stage primaryStage) {
+        MainWindow window = new MainWindow(new LogicManager());
```
ChinJunAn, line 23: I like how you separated the definition of the GUI components into another class. This would make it easier to for future developments. Also, it enforces OOP. Good job!

src/main/java/aqua/aquatask/AquaDeadline.java
```diff
@@ -0,0 +1,77 @@
+package aqua.aquatask;
+
+import java.time.LocalDateTime;
+import java.util.Optional;
```
ChinJunAn: I have noticed your usage of Optionals. I believe this is a good practice to prevent null exceptions. Also, this shows your competency in java programming

src/main/java/aqua/aquatask/AquaEvent.java
```diff
@@ -0,0 +1,93 @@
...
+import java.time.LocalDateTime;
+import java.util.Optional;
+
+import aqua.util.DateUtils;
+
+
+/** Implementation of AquaTask to represent an Event */
+public class AquaEvent extends AquaTask {
+    /** Tag of {@code from} argument when parsing. */
+    public static final String TAG_FROM = "from";
```
ChinJunAn: Great job following the coding standard!

src/main/java/aqua/logic/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,42 @@
...
+import aqua.aquatask.AquaDeadline;
+import aqua.exception.IllegalSyntaxException;
+import aqua.logic.ArgumentMap;
+import aqua.util.DateUtils;
+
+
+/**
+ * A full implementation of AddTaskCommand that creates and adds AquaDeadlines.
+ */
+public class AddDeadlineCommand extends AddTaskCommand {
```
ChinJunAn, line 12: Again, great job on practicing OOP by separating the command classes!

src/main/java/aqua/storage/Loadable.java
```diff
@@ -0,0 +1,14 @@
...
+
+
+/** Represents an object whose state can be loaded. */
+public interface Loadable {
+    /**
+     * Loads the the state of this object from a data stored in hard disk.
+     *
+     * @throws LoadException if an error occured while loading.
+     */
+    public void load() throws LoadException;
```
ChinJunAn, line 13: May I check the purpose of this class?

ChinJunAn (commented):
Very well-written code! I could learn from you

src/main/java/aqua/storage/Loadable.java
daitenshionyan: Thank you @ChinJunAn for spotting this. I think this class will only be implemented by `AppManager` throughout this whole iP.

At first I thought there might be more things to load like maybe a Config class that stores the user preference (font size, path to save directory, etc) or a LogRecord class that stores the log records and so on. So then all of these classes will implement this Loadable interface and stored into a `List` where we use a for loop or Streams to laod all of them.

```java
for (Loadable loadable : loadables) {
    loadable.load();
}
```

```java
loadables.stream().forEach(loadable -> loadable.load());
```

The `try` and `catch` block has been ommitted and `loadables` is a `List<Loadable>` that has been initialised else where (maybe the constructor).

I think the benifits of this over line by line calling the `load()` method is that we can handle the LoadException easier. For example, if we want the program to continue loading even if one class fails to load then we can do something like this.

```java
List<LoadException> loadExs = new ArrayList<>();
loadables.stream()
        .forEach(loadable -> {
                try {
                    loadable.load();
                } catch (LoadException loadEx) {
                    // just add to list and continue with next
                    loadExs.add(loadEx);
                }
            });
if (!loadExs.isEmpty()) {
    for (LoadException loadEx : loadExs) {
        loadEx.printStackTrace();
    }
} else {
    System.out.println("Loaded everything successfully");
}
```

Again, for the scale of the project I do not think these will be used.
Thank you for your comment!!

PR #58, opened 2023-01-21
[HJ-PRK] iP

hj-prk:


src/main/java/Duke.java
```diff
@@ -1,4 +1,4 @@
-import java.util.Scanner;
+import java.util.*;
```
jamieeeleow: Perhaps the imported classes could be listed explicitly?

src/main/java/Duke.java
```diff
@@ -1,5 +1,15 @@
 import java.util.*;
 public class Duke {
+    public static class task{
```
jamieeeleow: Maybe the class name could be ```Task``` instead of ```task``` with a small 't' to adhere to the coding standards where class names are written in PascalCase?

src/main/java/Duke.java
```diff
@@ -55,8 +61,33 @@ public static void main(String[] args) {
                 System.out.println("OK, I've marked this task as not done yet:");
                 System.out.println("[" +current.isDone+ "] " + current.todo);
                 System.out.println(line);
-            } else {
-                tasklist.add(new task(tasklist.size()+1, comm, " "));
+            } else if(comm.startsWith("todo")) {
+                String doit = comm.substring(5, comm.length());
```
jamieeeleow: Maybe the naming of the variable can be changed to camelcase ```doIt``` instead 😄

src/main/java/Duke.java
```diff
@@ -1,16 +1,22 @@
 import java.util.*;
 import java.lang.*;
+
+enum type {
```
jamieeeleow: I like the use of enum type!! 👍

src/main/java/Duke.java
```diff
@@ -85,7 +95,7 @@ public static void main(String[] args) {
                 } else if (comm.startsWith("deadline")) {
                     String doit = comm.substring(9, comm.length());
                     String[] parts = doit.split("/by");
```
jamieeeleow: 👍 use of the plural form for the list of String objects.

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task {
+
+    String from;
```
jamieeeleow: Perhaps modifiers could be added to the variables?

src/main/java/Duke.java
```diff
@@ -1,5 +1,8 @@
+import java.io.FileNotFoundException;
 import java.util.*;
 import java.lang.*;
+import java.io.FileWriter;
+import java.io.IOException;
```
jamieeeleow: Maybe the ordering of import classes can be a tad bit more consistent 😃

src/main/java/Week2/src/main/Duke.java
```diff
@@ -145,14 +145,19 @@ public static void main(String[] args) throws IOException {
                     lining();
                     writeOn(current);
 
+                } else if (comm.startsWith("find")) {
+                    String keyword = comm.substring(6, comm.length());
+                    Search sr = new Search(tasklist);
```
jamieeeleow: I like that the variable names are short, making the codes look neat and cleaner. However, the shortened names might be a bit confusing for different variables such as ```Scanner sc``` and ```Search sr``` since there is only one letter difference. Perhaps the names could be replaced with something more explicit? 😀

src/main/java/Week2/src/main/Deadline.java
```diff
@@ -1,12 +1,27 @@
 package Week2.src.main;
+
+/**
+ * A subclass of Task
```
jamieeeleow: Maybe some punctuations should be added at the end of each description line?

src/main/java/Week2/src/main/Ui.java
```diff
@@ -43,4 +34,18 @@ public void hello() {
...
         Duke.lining();
     }
+
+    public void bye() {
+        Duke.lining();
+        System.out.println("Bye. Hope to see you again soon!");
+        Duke.lining();
+    }
+
+    public void showEmptyError() {
```
jamieeeleow: I like the naming of the methods that makes it really clear as to what the function does.

PR #68, opened 2023-01-21
[ShanHng] iP

ShanHng:
### Babe

Babe frees your mind of having to remember things you need to do. It's,
* text-based
* easy to learn
* ~~FAST~~ _SUPER_ FAST TO USE

It's very simple to set up!
1. Download it 
2. Double-click it 
3. Start adding tasks!

The best part is it is **FREE**

Features:
- [x] Managing Tasks
- [x] Managing Deadlines
- [ ] Reminders (coming soon)

---

```java
public class Main {
    }
}

```


src/main/java/Babe.java
```diff
@@ -0,0 +1,448 @@
+import java.io.*;
```
heeeyi: I feel here more appropriate to import the specific packages instead of using whildcard import?

src/main/java/Babe.java
```diff
@@ -0,0 +1,448 @@
...
+
+    /**
+     * Length of user input.
+     */
+    private int userInputLen = 0;
+
+    /**
+     * List of Tasks Babe received from the user.
+     */
+    private ArrayList<Task> memory = new ArrayList<>();
```
heeeyi: Here more appropriate to name it as "memories" to show it a collection of items?

src/main/java/task/Task.java
```diff
@@ -0,0 +1,68 @@
+package task;
+
+/**
+ * Interface for list items in Babe,
+ */
+public abstract class Task {
+    /**
+     * String icon for marked.
+     */
+    static String MARKED = "[X]";
```
heeeyi: Consider using a common prefix for the static Strings?

src/main/java/Babe.java
```diff
@@ -0,0 +1,448 @@
...
+                                endDateIndex - 1));
+                        endDateAndTime = chatBot.formatDate(chatBot.rebuildUserInput(endDateIndex,
+                                inputLength));
+                        chatBot.addEvent(chatBot.rebuildUserInput(1, startDateIndex - 1),
+                                startDateAndTime,
+                                endDateAndTime,
+                                true);
+                        chatBot.save();
+
+                    } catch (Exception e) {
```
heeeyi: May be a better practice to catch the specific types of Exceptions, instead of catching the general Exception

src/main/java/Babe.java
```diff
@@ -0,0 +1,448 @@
...
+/**
+ * <h1> Hi Babe! </h1>
+ * Babe is a chat bot created as part of the individual project (iP) under the course CS2109T.
+ * It is a duplicate of Duke with some personal flair.
+ *
+ * @author Shan Hern Hng
+ * @version 1.0
+ * @since 17 January 2023
+ */
+public class Babe {
```
heeeyi: It is nice that you format the JavaDoc for author, version and time! I like it!

heeeyi (commented):
In generall, nice and clean codes that strictly follows the coding standard! Keep it up~

src/main/java/Babe.java
WindyCall: Aggreed! It would be better to explicitly list all classes imported.

src/main/java/babe/Babe.java
```diff
@@ -0,0 +1,86 @@
...
+ * @author Shan Hern Hng
+ * @version 1.0
+ * @since 17 January 2023
+ */
+public class Babe {
+
+    /**
+     * A TaskList object for this Babe.
+     */
+    private TaskList taskList = new TaskList();
```
WindyCall, line 22: Since you've arranged this class inside babe package, would it be better to use protected modifier?

src/main/java/babe/Babe.java
```diff
@@ -0,0 +1,86 @@
...
+                    Storage.save(chatBot.taskList);
+
+                } else if (parseOutputs.get(0).equals("delete")) {
+                    int index = Integer.parseInt(parseOutputs.get(1));
+                    chatBot.taskList.deleteTask(index);
+                    Storage.save(chatBot.taskList);
+
+                } else {
+                    throw new NonsenseInputException();
+                }
```
WindyCall: For this list of if and else, should using switch/case condition be with higher readability?

src/main/java/babe/Babe.java
```diff
@@ -0,0 +1,86 @@
...
+            } catch (Exception e) {
```
WindyCall: Is it better to catch specific exception instead of this kind of overall catching? There might be some problems if doing so(i.e catching some unexpected errors not handled properly)

src/main/java/babe/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    private ArrayList<Task> tasks = new ArrayList<>();
+
+    /**
+     * Adds a ToDo to memory.
+     * Calls the ToDo constructor and inserts created ToDo into this Babe's memory.
+     * @param content The description of the ToDo item.
+     */
+    public Task addToDo(String content, boolean toNotify) {
+        ToDo item = new ToDo(content);
```
WindyCall, line 35: Is the name item not suitable here? Maybe Task or todoTask make more sense?

src/main/java/babe/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Adds an Event to memory.
+     * Calls the Event constructor and inserts created Event into this Babe's memory.
+     * @param content The content of the Event item.
+     * @param startDate The start date of the Event. May include time too.
+     * @param endDate The end date of the Event. May include time too.
+     */
+    public Task addEvent(String content, String startDate, String endDate, boolean toNotify) {
```
WindyCall: Should the naming for boolean variable start from **is**, **has** etc?

WindyCall (commented):
Generally very good practice of naming and most of the codes have excellent code quality!!! Keep  it up and nice work!!

src/main/java/task/Task.java
ShanHng: That is a good point! Thank you!

PR #73, opened 2023-01-21
[Jnjy] iP

Jnjy:
# Chad

>“Your mind is for having ideas, not holding them.” – David Allen (source) 

Chad, never worry about missing your deadlines anymore. Ths solution lies in me. It's
- text-based
- ~~easy~~ EFFORTLESS to learn
- ~~fast~~ **_SUPER_** FAST to use 

<br/>

All you need todo is,
1. download it from [here](https://github.com/Jnjy/ip/releases/tag/v0.1)
2. double-click it
3. **Chad** with me
4. let me manage your tasks for you :fire:

<br/>

You're a Brokie? No worries, it is **FREE**!

<br/>

Features:
- [x] Managing tasks (Mark, Unmark, Delete)
- [x] Managing deadlines
- [ ] Search tasks
- [ ] Insane Graphic User Interface (Coming soon)

<br/>
If you a true Java programmer, you can use it to practice Java too. Here's the main method:
<br/>
<br/>

``` java
    new Duke().run();
}
```

<br/>


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,96 @@
...
+    public String getResponse(String request) {
+        try {
+            Command com = new Parser(request, tasks).processRequest();
+            String response = com.execute(tasks);
+            return response;
+        } catch (DukeException e) {
+            return e.getMessage();
+        }
+    }
+    /**
```
ScorpiusSigma: Do you reckon there should be a new line in between the lines to separate the methods?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,96 @@
...
+
+        while (!input.equalsIgnoreCase("bye")) {
+            Command com = new Parser(input, tasks).processRequest();
+            String response = com.execute(tasks);
+            UI.printRes(response);
+            input = sc.nextLine();
+        }
+
+        sc.close();
+    }
```
ScorpiusSigma: Do you reckon there should be a gap between 2 methods between line 72 to

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,59 @@
...
+        if (req.length < 2) {
+            throw new MissingArgumentException("☹ OOPS!!! You're missing the task description");
+        }
+
+        req = req[1].split("/by ");
+
+        // check presence of argument
+        if (req.length < 2) {
+            throw new MissingArgumentException("☹ OOPS!!! You're missing the task deadline");
+        }
```
ScorpiusSigma: How do you think this can be improved?

src/main/java/duke/command/AddEventCommand.java
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            return "Great! I've added this task for you \n" + newEvent
+                    + "\nYou have " + tasks.numOfTask() + " tasks in the list";
+        } catch (DateTimeException error) {
+            throw new InvalidArgumentException("Wrong date format! "
+                    + "Please follow the format YYYY-MM-DD (e.g. 2000-01-01)");
+        }
+    }
+}
```
ScorpiusSigma: I'm not sure I like how long this is. Perhaps you can split the exception check to be in another method?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Constructor method to create find command.
+     * @param request User's request to find duke.task.
+     */
+    public FindCommand(String request) {
+        this.request = request;
+    }
+
+    @Override
+    public String execute(TaskList tasks) {
```
ScorpiusSigma: Do you reckon adding javadocs to all your execute methods would allow the user to make sense of what that method does for that class?  This can be applied to other execute methods in other classes :)

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,18 @@
...
+import duke.storage.TaskList;
+
+/**
+ * Class use to handle command: view task list.
+ * Allows user to view the task list.
+ */
+public class ListCommand extends Command {
+    @Override
+    public String execute(TaskList tasks) {
+        // if there is no task in the list
```
ScorpiusSigma: I like how you have have inline comments which makes reading even better!

ScorpiusSigma (commented):
Overall, I like how you document your code and implementation. However, I felt some methods can be abstracted even further such as having another method to check for errors. Good Job! 👍

src/main/java/duke/command/AddTodoCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public String execute(TaskList tasks) throws MissingArgumentException {
+
+        String[] task = request.strip().split("todo");
+
+        // check missing description of todo task
+        if (task.length == 0) {
+            throw new MissingArgumentException("☹ OOPS!!! The description of a todo cannot be empty.");
```
securespider: Instead of having this message as parameter in your constructor, if this is the default string for this exception it might be better to put it inside the exception itself

src/main/java/duke/command/FindCommand.java
securespider: I agree i think doing so makes it more explicit and more clear for developers to understand what the execute does for each of the commands

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,70 @@
...
+import duke.command.AddDeadlineCommand;
+import duke.command.AddEventCommand;
+import duke.command.AddTodoCommand;
+import duke.command.Command;
+import duke.command.ExitCommand;
+import duke.command.FindCommand;
+import duke.command.ListCommand;
+import duke.command.MarkCommand;
+import duke.command.RemoveCommand;
+import duke.command.UnmarkCommand;
```
securespider: Good job! Each package is declared explicitly instead of as in a wildcard import

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param tasks duke.task array to store the tasks added by the user
+     */
+    public Parser(String request, TaskList tasks) {
+        this.request = request;
+        this.tasks = tasks;
+    }
+
+    /**
+     * Function to process the user's request and get the response
+     * @return
```
securespider: Remember to specify what the method is returning

src/main/java/duke/storage/LocalStorage.java
```diff
@@ -0,0 +1,143 @@
...
+        file.getParentFile().mkdirs();
+
+        if (file.exists()) {
+            return file;
+        }
+
+        try {
+            file.createNewFile();
+        } catch (IOException io_error) {
+            io_error.printStackTrace();
```
securespider: Might not be good for the stack trace to be printed like this. Maybe you could reformat the error message to be nicer for the user to read? Otherwise, its good that it is verbose and the user knows that there is an error there

src/main/java/duke/storage/LocalStorage.java
```diff
@@ -0,0 +1,143 @@
...
+                        try {
+                            LocalDate formattedDueDate = LocalDate.parse(dueDate);
+                            Deadline deadline = new Deadline(taskDesc, formattedDueDate);
+                            if (taskStatus.equals("1")) {
+                                deadline.markComplete();
+                            }
+                            tasks.add(deadline);
+                            break;
+                        } catch (DateTimeException error) {
+                            throw new InvalidArgumentException("Wrong date format! Please follow the format "
```
securespider: Similar to above, since this string is used in multiple lines for the same exception, might be good to store it in the exception itself

securespider (commented):
Good job!
Minor changes to the code can be done to make it look better and clearer for other developers.

PR #79, opened 2023-01-21
[Anunaya Joshi] IP

anunayajoshi:
### This PR consists of changes to include **CRUD** functionality and _Unit Testing_ to the application. 😄 

The changes made are as follows:

- Ability to greet and exit 
- Ability to add, list, delete todos, deadlines and events 
- Automated Text UI Testing as well as JUnit
- Saving to hard disk
- Marking and unmarking tasks as done or not.

The things I struggled with while working on this application were:

1. Following OOP best practices and deciding on the classes to use. 
2. Using git and the branches effectively and dealing with merge conflicts. 

Additional things that need to be done are:

- [ ] Creating a GUI for the application
- [ ] Adding in checkstyle functionality 

If you would like to view work done by me, you can visit by github [here](www.github.com/anunayajoshi)

One quote I like is

> Spend more time and do it once, instead of doing it multiple times. 

A sample input for the application is like `mark 3` 

Example code function is as follows:

```java
public void setTaskStatus(int index, boolean isDone) {
        Task task = this.tasks.get(index - 1);
        task.setDone(isDone);
        Ui.separator();
        System.out.println("\tOk, I have marked this task as " + (isDone ? "done" : "not done yet") + ":\n\t\t"
                + task);
        Ui.separator();
    }


```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+    private final Storage storage;
+
+    public Parser(TaskList tasklist, Ui ui, Storage storage) {
+        this.tasklist = tasklist;
+        this.ui = ui;
+        this.storage = storage;
+    }
+
+    /**
+     * Takes in user's input and obtains the first word
```
huixuant: Perhaps the format of the JavaDoc header comment here can be further improved? The coding standard states that the first sentence should be written as a **short summary of the method**, as Javadoc automatically places it in the method summary table (and index). However, the first line of this header comment appears to be truncated and not complete. 

Additionally, perhaps consider adding an empty line between the description and parameter section, as per the coding standard? This applies to all other JavaDoc comments without the empty line.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+     * and uses it to run different operations,
+     * and then uses the rest of the words as arguments for the operation.
+     * @param input User's input
+     */
+    public String parse(String input) {
+        String str;
+        try {
+            String[] arr = input.split(" ", 2);
+            boolean details = arr.length != 1;
```
huixuant: Perhaps consider using another name for the boolean variable `details`? According to the coding standard, boolean variables should be named to sound like booleans. A possible alternative could be something like `hasDetails`. 😄

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,36 @@
+package duke;
+
+/**
+ * Task which has 3 variations: ToDo, Deadline and Event.
+ * Can be Marked to be completed or not completed.
+ */
+public class Task {
+    private final String description;
+    private Boolean done;
```
huixuant: Similar to one of the comments above, perhaps consider renaming `done` to `isDone` so that it complies with the coding standard?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,51 @@
...
+    public Ui() {
+        greet();
+        sc = new Scanner(System.in);
+    }
+
+    public boolean hasUserInput() {
+        return sc.hasNextLine();
+    }
+
+    public String userInput() {
```
huixuant: Perhaps consider renaming this method to start with a verb so that it complies with the coding standard? For example, a possible alternative could be `getUserInput()`.

src/test/java/duke/EventTest.java
```diff
@@ -0,0 +1,24 @@
...
+    @Test
+    public void dummyTest(){
+        assertEquals(2, 2);
+    }
+
+    /**
+     * Tests if the event toString method prints correctly.
+     */
+    @Test
+    public void eventToStringTest(){
```
huixuant: Perhaps the test method could be renamed such that it follows the format given in the coding standard `featureUnderTest_testScenario_expectedBehavior()`?

huixuant (commented):
Code seems well-organised with elements of OOP being implemented. Other than minor coding standard violations, code is generally readable and easy to understand with proper naming in place.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class Deadline extends Task {
+
+    protected LocalDate by;
+
+    /**
+     * Creates a Deadline Task
+     *
+     * @param description in String format.
+     * @param by          Which is a LocalDate.
```
MrTwit99: Is there perhaps some additional space indented error or there was something here but I can't view it?
```suggestion
     * @param by Which is a LocalDate.
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+                break;
+            default:
+                throw new DukeException("Sorry, I don't know what that means.");
+            }
+
+            // handle changes to arraylist
+            storage.writeFile(tasklist);
+        } catch (DukeException e) {
+            str = ui.showError(e);
+
```
MrTwit99: There might be an additional space that was accidentally entered. Not big of an issue.

```suggestion
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,91 @@
+package duke;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
```
MrTwit99, line 8: I was just wondering, would it be slightly better to leave a line to separate the imports from diff packages iirc 2030S used to do this... Not sure if its needed here.

```suggestion
import java.io.IOException;

import java.util.Scanner;

```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+        String list = "";
+        for (int i = 0; i < this.tasks.size(); i++) {
+            int currentNumber = i + 1;
+            Task task = this.tasks.get(i);
+
+            list = list + "\t" + currentNumber + "." + task + "\n";
+        }
+        return list;
+    }
+
```
MrTwit99: Would be great if you could add the JavaDocs in when you have the time.

MrTwit99 (commented):
Overall, there isn't much issues besides some naming issues as mentioned by another reviewer. Great job with the iP project and you could perhaps review through the suggestions I have made.

src/main/java/duke/Storage.java
MrTwit99: Edit to above comment:

After running the checkstyle feature, I think the above suggestion is invalid for most ``import`` situations! Do take note when there's a need for it.

PR #83, opened 2023-01-22
[Leon Lim] iP

Tempura-Person:
# Duke (With Less User Respect)

> "Duke." - Duke (2023)

## What's Duke (With Less User Respect)?
Duke has **had enough** with users ordering it around, so it is now in its _rebellious_ phase.
It still will do what you say, but now it's:
* less **polite**
* more **sassy** 🔥 
* secretly **cursing** you behind your back...?

## Installation
1. Download **Duke (With Less User Respect)** [here](https://github.com/Tempura-Person/ip/releases/tag/A-Jar)
2. **???**
3. **Profit!**

## Maintenance Notes
- [x] Add code 
- [ ] Get good
- [ ] Remove Herobrine

## Safety Advisory
Please remember to talk with your Duke (With Less User Respect) every day, or it will run
``` *.java
while (true) {
    System.out.println("Duke.\n");
}
```
and cause your OS to brick upon startup.


src/main/java/features/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class Storage {
+
+    /**
+     * Loads taskList from data/duke.txt, if any.
+     * @return the loaded taskList.
+     */
+    public TaskList loadTaskList() {
+
+        TaskList toLoad = new TaskList();
```
astraxq: `toLoad` sounds a bit vague on what this variable does. `newTaskList` for example sounds more precise.

src/main/java/features/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        } catch (IOException err) {
+            System.out.println("Sorry, could not save that.");
+        }
+
+        try {
+            Scanner fileScan = new Scanner(dukeSave);
+            while (fileScan.hasNextLine()) {
+                String toEval = fileScan.nextLine();
```
astraxq: Similar case, `toEval` can be rephrased as a noun.

src/main/java/features/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                String[] evalArray = toEval.split("=");
+                switch (evalArray[0]) {
+                case ("T"):
+                    ToDo addToDo = new ToDo(evalArray[2]);
```
astraxq: `addToDo` should be a noun. (e.g `toDo` / `toDoTask`), like how you name the task that you are about to delete as `deletedTask`. 

Applies to `addDeadline` and `addEvent`.

src/main/java/features/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+        return toLoad;
+    }
+
+    /**
+     * Saves taskList to data/duke.txt
+     * @param toSave The TaskList object to save.
+     * @throws IOException  If TaskList cannot be saved.
+     */
+    public void saveTaskList(TaskList toSave) throws IOException {
+        FileWriter dukeWrite = new FileWriter("data/duke.txt");
```
astraxq: Noun variable naming convention (`dukeWrite` -> `dukeWriter` e.g)

src/main/java/features/TaskList.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.ArrayList;
+
+import tasks.Task;
+
+/**
+ * Contains and modifies tasks entered by the user.
+ */
+public class TaskList {
+    protected ArrayList<Task> list;
```
astraxq: Plural form should be used for a collection of objects. 

E.g
`int[] values;`

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,65 @@
...
+        try {
+            this.fromDateTime = LocalDateTime.parse(from, DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm"));
+        } catch (DateTimeParseException err) {
+            this.fromDateTime = null;
+        }
+        try {
+            this.toDateTime = LocalDateTime.parse(to, DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm"));
+        } catch (DateTimeParseException err) {
+            this.toDateTime = null;
+        }
```
astraxq, line 35: Instead of using two try catch, throw a `DukeException` to handle both cases. 
Having an event from: `LocalDateTime` to: `null`, kind of defeats the purpose of the event task?

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+        } catch (DateTimeParseException err) {
+            this.byDateTime = null;
+        }
+    }
+
+    public String getByTime() {
+        if (this.byDateTime != null) {
+            return this.byDateTime.format(DateTimeFormatter.ofPattern("MMM dd yyyy, HHmm")) + " hrs";
+        } else {
+            return this.by;
```
astraxq, line 33: This `else` case is basically handling the null value you assigned to `byDateTime`,  would strongly suggest you throw a DukeException to handle invalid date format.

astraxq (commented):
Overall great job! Code is written very neatly, just some minor changes I've suggested. Good luck on the subsequent taskings!

src/main/java/features/Storage.java
Tempura-Person: Thanks, i've fixed this now.

src/main/java/features/Storage.java
Tempura-Person: Same as above, i've made the change now.

src/main/java/features/Storage.java
Tempura-Person: Same as above, i've made the change now.

src/main/java/features/Storage.java
Tempura-Person: Same as above, i've made the change now.

src/main/java/features/TaskList.java
Tempura-Person: Same as above, i've made the change now.

src/main/java/tasks/Event.java
Tempura-Person: I feel that a DukeException for an invalid date/time format would be very limiting for users who might want to put an abstract representation of time (e.g. a person may want to do event X from 'after lunch' to 'before dinner')
What are your thoughts on this?

astraxq: I've actually never thought of that, cool reasoning for it!

src/main/java/Duke.java
```diff
@@ -20,207 +20,196 @@ public static void main(String[] args) {
         boolean LoopEnd = false;
         while (!LoopEnd) {
```
ChangGittyHub: I like how your code has no arrowhead despite a very long code! keep it up:)

src/main/java/Duke.java
```diff
@@ -20,207 +20,196 @@ public static void main(String[] args) {
...
-                            System.out.println("\n" + border + "[ERROR]\nUh, mark command format is used wrongly.\nCorrect format is as follows:\n" +
-                                    "[ mark <insert INTEGER> ]\n" + border);
+                    // Duke lists out all Task names in TaskList when input is "list"
+                    case ("list"):
+                        // ERROR: list format is anything other than [ list ]
+                        if (UserScan.nextLine().length()>0) {
+                            throw new DukeException("\n" + border + "[ERROR]\nUh, list command format is used wrongly.\nCorrect format is as follows:\n" +
+                                    "[ list ]\n" + border);
+                        }
+                        if (TaskList.size()==0) {
```
ChangGittyHub: could this be abstracted out to TaskList.isEmpty for it be more meaningful? just a suggestion!

src/main/java/DukeException.java
```diff
@@ -1,5 +1,12 @@
 public class DukeException extends Exception{
+    protected String ErrorMessage;
+
     public DukeException(String ErrorMessage) {
         super(ErrorMessage);
+        this.ErrorMessage = ErrorMessage;
+    }
+
+    public String PrintErrorMessage () {
```
ChangGittyHub: maybe camel casing should be done here?

src/main/java/Parser.java
```diff
@@ -1,54 +1,70 @@
+import Features.DukeException;
+import Features.TaskList;
+import Features.Ui;
+import UserCommands.*;
```
ChangGittyHub: maybe you can avoid * imports?

src/main/java/Features/Storage.java
```diff
@@ -24,27 +24,27 @@ public TaskList loadTaskList() {
...
-                            toLoad.add(addDeadline);
-                            break;
-                        case("E"):
-                            Event addEvent = new Event(evalArray[2], evalArray[3], evalArray[4]);
-                            if (evalArray[1].equals("X")) {
-                                addEvent.MarkDone();
-                            }
-                            toLoad.add(addEvent);
-                            break;
+                    case ("T"):
```
ChangGittyHub: I like how you fixed the indentation for switch case, a common mistake!

src/main/java/Features/Storage.java
```diff
@@ -24,27 +24,27 @@ public TaskList loadTaskList() {
...
+                        ToDo addToDo = new ToDo(evalArray[2]);
+                        if (evalArray[1].equals("X")) {
```
ChangGittyHub: maybe evalArray[1] can be abstracted out with a  meaningful name?

ChangGittyHub (commented):
sweet! great well written code:)

PR #84, opened 2023-01-22
[yitong241] iP

yitong241:
# DUKE
> "Just use DUKE"      -yitong

DUKE is:
- ~~easy~~ SUPER easy to learn 😎
- easy to use

**Download** it [here](https://github.com/yitong241/ip).

Supported *operations*:
1. add
2. delete
3. list
4. mark as done
5. mark as undone

You can run the GUI with `Launcher`:
```java
public class Launcher {
    }
}
```

Features:
- [x] todo
- [x] deadline
- [x] event


src/main/java/Duke/Interpreter.java
```diff
@@ -0,0 +1,247 @@
...
+     * The method mark marks the Task as done
+     * @param command the command by the user
+     * @param table the task table
+     * @return Task
+     * @throws InvalidCommandException
+     * @throws OutRangeException
+     */
+    public static Task mark(String command, TaskTable table) throws InvalidCommandException, OutRangeException {
+        int len = table.size();
+        int index;
```
NoyaRoeT, line 30: The Java coding standard says to initialize variables where they are declared. Would it be better to declare index at line 39 instead?

src/main/java/Duke/Tasks/Disk.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Scanner;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalDateTime;
+
+public class Disk {
+
+    private final String path;
+    private final DateTimeFormatter format = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
```
NoyaRoeT, line 15: Perhaps using SCREAMING_SNAKE_CASE for constant names would better match the Java coding standards?

src/main/java/Duke/Tasks/Deadline.java
```diff
@@ -0,0 +1,78 @@
+package Duke.Tasks;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents the Deadline class
```
NoyaRoeT, line 7: Would it be redundant to say that the Deadline class represents the Deadline class? I noticed the same thing in your other classes as well.

src/main/java/Duke/Tasks/Event.java
```diff
@@ -0,0 +1,88 @@
...
+public class Event extends Task {
+    protected LocalDateTime startTime;
+    protected LocalDateTime endTime;
+
+    /**
+     * The constructor of Event
+     * @param desc the description of event
+     * @param startTime the start time of event
+     * @param endTime the end time of event
+     * @param done whether the event is done
```
NoyaRoeT, line 19: A very small thing, but Java standards suggest adding punctuation behind each parameter description.

NoyaRoeT (commented):
Overall, I found your code easy to read for the most part. I noted a few minor coding standard violations and some of the header comments also seem redundant.

PR #93, opened 2023-01-22
[LimYanLing] iP

yanlings:

- text-based
- easy to learn
-  ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from here.
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [X] Managing tasks
- [ ] Reminders (coming soon)

```
public class Launcher{
    public static void main(String[] args) throws DukeException, IOException {
        Application.launch(Main.class,args);
    }
}
```


src/main/java/duke/handleInvalidArgs.java
```diff
@@ -0,0 +1,72 @@
+package duke;
+
+
+/**
+ * This is a class to handle the invalid number of arguments for each of the class
+ */
+public class handleInvalidArgs {
```
Emrysil: Perhaps can rename to InvalidArgsHandler?since we're suppose to use nouns for class names:)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,74 @@
+package duke;
+
+import java.io.*;
```
Emrysil, line 3: Maybe try to specify the class being imported instead of just importing *

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,56 @@
+package duke;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+/**
+ * This is a class for Event action
+ */
+import duke.Task;
+public class Event extends Task {
```
Emrysil, line 8: I think import statements should be grouped together

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,56 @@
...
+    protected LocalDateTime end;
+
+    /**
+     * This is a constructor for event class
+     * @param description description of event
+     * @param date date during which the event happens
+     */
+    public Event(String description, String date)  {
+        super(description);
+        String[] fromandto = date.split("/");
```
Emrysil: Maybe rename to fromAndTo to following the camelCase naming convention

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,56 @@
...
+        return "[E]" + super.toString() + " (from:" + from + "to:" + to + ")";
+    }
+
+    /**
+     * Returns a string representation of the Event object in the list
+     * @return a string representation of the Event object in the list
+     */
+    @Override
+    public String toSaveString() {
+        String timeStr = this.start.format(DateTimeFormatter.ofPattern("HH:mm, MMM dd yyyy"));
```
Emrysil, line 51: Maybe rename to timeStart? timeStr can be a bit misleading as it tends to be interpreted as timeString

src/test/java/EventTest.java
```diff
@@ -0,0 +1,15 @@
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
```
Emrysil, line 3: I might be wrong but I think you don't have to separate the two imports

PR #105, opened 2023-01-22
kennethk-1201 iP

kennethk-1201:
# Duke

Duke is a **task application** :open_book: where you can:
- Add ToDo, Deadline and Event tasks
- Delete your tasks
- Filter your tasks based on the time

How to set up?
1. Download the release from the repo.
2. Run it.
3. Add your tasks!

Current releases:
- [x] v1.0
- [ ] v1.1
- [ ] v1.2

If you're a Java developer, here's the `main` method:

```java
    new Duke("./tasks.txt").run();
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,32 @@
+import sys.Storage;
```
CloudHill, line 3: Shouldn't `Duke` be in a package?

The coding standard states that "Every class should be part of some package."

Maybe you can put all your code under a `duke` package.

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Represents a command to add a task to the task list.
+ */
+public class AddCommand extends Command {
+    private String input;
+
+    /**
+     * Constructor for AddCommand
+     * @param input Input to parse.
```
CloudHill, line 22: Looks like you're missing an empty line between the description and parameters, I've noticed this in other constructors as well.

Plus, constructors are technically methods, so I think the header should still start with a verb, something like `Constructs a new AddCommand.`

src/main/java/command/Command.java
```diff
@@ -0,0 +1,42 @@
...
+ * Represents the command given by the user.
+ */
+public abstract class Command {
+    private String regex;
+
+    /**
+     * Constructor for Command
+     * @param regex Regex for command argument to match with.
+     */
+    Command (String regex) {
```
CloudHill: Looks like you have an extra space here before the parenthesis.

src/main/java/sys/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Converts a user input into a Command object.
+     *
+     * @param input The input line given by the user.
+     * @return Command object representing the input.
+     */
+    public Command parse(String input) {
+        // Handle control flow
+        switch (input) {
+            case "bye":
```
CloudHill: Based on the coding standards, `case` clauses shouldn't be indented (yeah, it's kinda weird).

So, it should look something like:

```java
switch (x) {
case 1:
    // ...
    break;
case 2:
    // ...
    break;
// ...
}
```

src/main/java/helper/DateTimeHelper.java
```diff
@@ -0,0 +1,126 @@
...
+        Integer month = Integer.valueOf(values[1]);
+
+        String[] yearAndTime = values[2].split(" ");
+        Integer year = Integer.valueOf(yearAndTime[0]);
+        Integer hour = Integer.valueOf(yearAndTime[1].substring(0, 2));
+        Integer minute = Integer.valueOf(yearAndTime[1].substring(2));
+
+        // Convert the time units into a LocalDateTime object.
+        try {
+            return LocalDateTime.of(year, month, day, hour, minute);
```
CloudHill, line 44: Unrelated to coding standards, but you can check out `LocalDateTime.parse`

You can pass it the specific format you want and it'll also throw a `DateTimeParseException` if the string is in the wrong format, so you can save the trouble of validating it yourself.

[Here's the docs](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#parse-java.lang.CharSequence-java.time.format.DateTimeFormatter-) if you want to read about it.

CloudHill (commented):
Overall, LGTM, just a few nits related to the coding standards. It's pretty impressive that you've commented pretty much all the public methods and classes.

src/main/java/Duke.java
```diff
@@ -1,10 +1,35 @@
...
+ * Represents the whole Duke program.
+ */
 public class Duke {
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
```
Bisceto, line 8: This can be encapsulated in the UI class

src/main/java/Main.java
```diff
@@ -0,0 +1,29 @@
...
+    public void start(Stage stage) {
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
+            AnchorPane ap = fxmlLoader.load();
+            Scene scene = new Scene(ap);
+            stage.setScene(scene);
+            fxmlLoader.<MainWindow>getController().setDuke(duke);
+            stage.show();
+        } catch (IOException e) {
+            e.printStackTrace();
```
Bisceto, line 26: Can have a more meaningful catch block to handle errors

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,50 @@
+package command;
```
Bisceto, line 1: Package should be in such a way that duke is still the root package, so it should be something like duke.command instead

src/main/java/command/Command.java
```diff
@@ -0,0 +1,45 @@
...
+        this.regex = regex;
+    }
+
+    /**
+     * Executes the command.
+     *
+     * @param tl the current list of tasks
+     * @param ui the user interface running.
+     * @param storage the storage location for the program.
+     * @throws DukeException If an invalid input is given.
```
Bisceto, line 34: Should have a return param statement for Javadoc

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Override
+    public Response execute(TaskList tl, Ui ui, Storage storage) throws DukeException {
+        // Find index of task to delete
+        Integer idx = Integer.valueOf(input.split(" ")[1]) - 1;
+
+        // Verify if task number is invalid:
+        if (idx < 0 || idx >= tl.numberOfTasks()) {
+            throw new InvalidCommandInputException("Task number is invalid!", "delete");
+        }
```
Bisceto, line 46: There is an IndexOutOfBounds Exception that you can consider using instead.

src/main/java/helper/DateTimeHelper.java
Bisceto: Agreed, you can remove alot of this code that converts your format into a LocalDateTime Object and just use their parse method along with DateTimeFormatters.

src/main/java/sys/Parser.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Represents a parser used to decode user input.
+ */
+public class Parser {
+
+    /**
+     * Constructor for Parser.
+     */
+    Parser() {};
```
Bisceto, line 22: No need to include this if you don't need a constructor: Java will have a default constructor even if you don't include it.

src/main/java/sys/Parser.java
```diff
@@ -0,0 +1,54 @@
...
+            } else if (input.matches("mark \\d+")) {
+                return new MarkCommand(input);
+            } else if (input.matches("unmark \\d+")) {
+                return new UnmarkCommand(input);
+            } else if (input.matches("delete \\d+")) {
+                return new DeleteCommand(input);
+            } else {
+                return new AddCommand(input);
+            }
+        }
```
Bisceto, line 52: Can consider using switch statements for cleaner code.

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,53 @@
+package task;
+
+import exception.InvalidDateFormatException;
+
+import helper.DateTimeHelper;
+
+import java.time.LocalDateTime;
```
Bisceto, line 9: Not too sure about the order of import statements here: But IIRC java imports should be at the top.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,147 @@
...
+
+            String start = content.substring(startIndex, endIndex - 5);
+            String end = content.substring(endIndex, content.length() - 1);
+            String eventContent = content.substring(0, startIndex - 8);
+
+            // Return event object.
+            return new Event(eventContent, isMarked, start, end);
+        default:
+            throw new InvalidTaskStringException("Unknown task type!");
+        }
```
Bisceto, line 112: This method feels very cluttered despite the use of switch statements. Perhaps you can consider moving the error handling for invalid tasks into their constructors! This will make this method much cleaner by removing all the error handling.

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,222 @@
...
+        return tasks.size();
+    }
+
+    /**
+     * Print the tasks that occurs on a given date. It is used for the occurs command.
+     *
+     * @param input The date to check.
+     * @throws InvalidDateFormatException If the date given does not follow the specified format.
+     */
+    public TaskList findTasksOnDate(String input) throws InvalidDateFormatException {
```
Bisceto, line 199: Might be more scalable if you catch the InvalidDateFormatException in this method and throw a DukeException instead

Bisceto (commented):
Good job with the javadocs! LGTM, just some nits to fix, such as refactoring the packages.

PR #107, opened 2023-01-22
[Andre Lin HuiKai] iP

4ndrelim:
# MyDuke
> _"Building the damned plane as we fly it"_ - __lima_dingdong__ [[look up](https://github.com/4ndrelim)]

__MyDuke__ ensures you are on track in completing your tasks so your plane do not crash and burn! It is:
- ***COMPLETELY FREE***
- text-based
- easy to learn
- reliable
- __MOST IMPORTANTLY__, ~~FAST~~ _SUPER FAST_ to use

To use, simply,
1. [git clone](https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository) from [here](https://github.com/4ndrelim/ip)
2. Navigate to `build/libs`
3. run `java -jar Duke.jar` 
4. add your tasks
5. let it manage! 😄 

__Features__:
- [x] Managing tasks
- [x] Saving tasks
- [x] Searching tasks
- [ ] Ordering tasks (_Coming soon_)

If you are a Java programmer, feel free to examine and improve the current product! Here's the `main` method:
```
public class Main {
        Duke duke = new Duke();
        duke.run();
    }
}
```


src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,73 @@
...
+    public String encode() {
+        return String.format("%s ### %s", this.isDone, this.description);
+    }
+
+    /**
+     * Checks if a task falls on a given date.
+     * @param date Date to check against.
+     * @return boolean value that checks if the task falls on the specified date.
+     */
+    public boolean fallsOnDate(LocalDate date) {
```
YuanXun9834, line 73: Should the method name be changed to one that sounds more like a boolean? 👍

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     *
+     * @param command
+     * @param echo The user input to be parsed to be used in the execution of its associated command.
+     * @return a boolean value that determines whether the program will continue or halt.
+     * @throws DukeException DukeException thrown if there is missing required information in the user's input.
+     */
+    public boolean execute(Command command, String echo) throws DukeException {
```
YuanXun9834: Should the various types of commands be abstracted out into different classes, so as to make use of dynamic binding?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+import duke.exceptions.DukeException;
+
+/**
+ * Duke is a personal assistant bot that helps user to track tasks.
+ *
+ * @author Andre Lin HuiKai
+ * @version CS2103T AY22/23 Semester 2
+ */
+public class Duke {
+    private FileManagement fileManager;
```
YuanXun9834, line 18: Great job in following the coding standard for your duke class ! 👍

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public void unmarkTask() {
+        this.isDone = false;
+    }
+
+    /**
+     * Displays formatted information about the task.
+     * @return String containing common information about a task.
+     */
+    public String getStatusIcon() {
```
YuanXun9834, line 48: I feel this method description is unnecessary as the class is already declared as an abstract class.

YuanXun9834 (commented):
Overall great job in following the coding standard in your iP. I've provided a few minor feedback points and hope it helps :)

src/main/java/duke/task/Task.java
4ndrelim: I think this already suggests a boolean? The method tells us whether a task indeed falls on a given date. Trying to intentionally add the conventional "is" / "can" / "has" sounds a little too forced and grammatically-off. Unless you have a better name for it in mind? haha

src/main/java/duke/command/Parser.java
4ndrelim: Yes for extensibility. But i decided against it because 
1. takes time 
2. The initial parsing using regex differs for each command and i can't see much commonalities among the commands at least up till the current context.
Thanks!

src/main/java/duke/Duke.java
4ndrelim: Thanks!

src/main/java/duke/task/Task.java
4ndrelim: Ah okay noted on this point

PR #111, opened 2023-01-23
[JavonTeo] iP

JavonTeo:
# Duke

> My brain is like the Bermuda Triangle. Information goes in and then it is never found again - Unknown ([source](https://www.epicquotes.com/my-brain-is-like-the-bermuda-triangle-information-goes-in-and-then-it-is-never-found-again/))

Are you a forgetful person? Ever experienced that awkward situation where you said yes to a family event, but you forgot that you had promised your boyfriend/girlfriend to bring he/her on a date on the same day, so now your mom is arguing with your partner??? Well, Duke is here for you!

- **text-based**
- **easy to learn**
- **~~FAST~~ _SUPER_ FAST to use**

All you need to do is,
2. double-click it.
3. add your tasks.
5. And it is FREE!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

### It can be used either:
1. via a text-based UI, or
2. via a GUI

To use the program via the text-based UI, run the `main` method in `Duke`:
```java
public class Duke {
        new Duke().run();
    }
}
```

To use the program via the GUI, run the `main` method in `Launcher`:
```java
public class Launcher {
    }
}
```


src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+    public void markUnmark(boolean marked) {
+        this.isDone = marked;
+        if (marked) {
+            System.out.println("Nice! I've marked this task as done:");
+        } else {
+            System.out.println("OK, I've marked this task as not done yet:");
+        }
+        System.out.println(this);
+    }
+}
```
tituswe: Have you considered implementing a `toString()` method for this parent class? This way you need not repeat the done status or the description in each of the child classes of `Task`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,165 @@
...
+                    }
+                    case "event" -> {
+                        Duke.addEvent(taskList, restOfString);
+                        taskListChanged = true;
+                    }
+                    case "delete" -> {
+                        Duke.delete(taskList, restOfString);
+                        taskListChanged = true;
+                    }
+                    default -> throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
tituswe: According to the [CS2103T coding standard](https://se-education.org/guides/conventions/java/intermediate.html) switch statements should not have indentation for the `case` clauses. Slight amendment needed here 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,165 @@
...
+            System.out.println(removedTask);
+            System.out.println("Now you have " + taskList.size() + " tasks in the list.");
+        } catch (NumberFormatException err) {
+            throw new DukeException("☹ OOPS!!! " + restOfString + " is not a valid integer for indexing the task list.");
+        } catch (IndexOutOfBoundsException err) {
+            throw new DukeException("☹ OOPS!!! There are less than " + restOfString + " tasks.");
+        }
+    }
+
 }
```
tituswe: You could consider using inheritance and polymorphism for your commands, this way you wouldn't have to keep adding new cases for new commands in your `main` code. On top of this your `main` class would be a lot cleaner!

src/main/java/ToDo.java
```diff
@@ -0,0 +1,17 @@
+public class ToDo extends Task {
+    public ToDo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.getStatusIcon() + " " + super.getDescription();
```
tituswe: I like the use of the `getStatusIcon()` and `getDescription()` methods for overriding the child class `toString()` method, however you could consider implementing the `toString()` method for the `Task` class as its a more conventional way to do it and it gets the job done just as well!

tituswe (commented):
Looks good! A few coding standard issues and suggestions that you could use. All the best!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,55 @@
...
+                c = new AddCommand(command, restOfString);
+                break;
+            }
+            case "delete": {
+                c = new DeleteCommand(restOfString);
+                break;
+            }
+            default:
+                throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
```
swx0: Don't need to indent for cases/default within the switch

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,38 @@
...
+        System.out.println("_______________________");
+    }
+
+    public void showError(String err) {
+        System.out.println(err);
+    }
+
+    public void show(String msg) {
+        System.out.println(msg);
+    }
```
swx0, line 61: Perhaps an additional class could be used to manage the logic of printing output by the various Commands. Instead of having to use the `show(String msg)` function each time an output line is displayed.

swx0 (commented):
Overall good. Just some suggestions to consider.

PR #123, opened 2023-01-23
[eugenetangkj] iP

eugenetangkj:
BooBot helps you to keep track of your tasks, saving time and ensuring that you are on track. Problems would be scared away with good task management provided by BooBot.

Main features of BooBot:
- Text-based UI
- Three types of tasks: _Todo_, _Deadline_ and _Event_
- Add and delete tasks from task list.

To get started with BooBot:
1. Install Java.
2. Download the source code [here](https://github.com/nus-cs2103-AY2223S2/ip).
3.  Run the BooBot application.

It takes only 3 steps! Isn't it easy? And not to mention, BooBot is entirely **free**! 😆 

Also, if you are a Java programmer, you can even use the source code to customise BooBot according to your needs! Here is the 
`main` method:
```java
public class Main {
    }
}
```
What are you waiting for? Download BooBot now!
- [x] Download BooBot
- [x] Save your tasks
- [ ] Say bye to your worries!


src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Represents a "bye" command that is entered by the user.
+ */
+public class ByeCommand extends Command {
+    private static final String EXIT_MESSAGE = "Goodbye. Hope that I have managed to scare all your problems away."
+            + "\nHave a great day! :)";
+
+
+    /**
+     * Constructs a <code>ByeCommand</code>.
```
VietAnh1010: You can instead use `{@code ByeCommand}` - it will be more concise.

src/main/java/duke/datetime/DateTime.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Determines if one date is equal to another date, based on year, month and day.
+     *
+     * @param start The Temporal encapsulating the start date and time.
+     * @param end The Temporal encapsulating the end date and time.
+     * @return true if both refer to the same day, else false.
+     */
+    public static boolean isEqualDate(Temporal start, Temporal end) {
```
VietAnh1010: Good naming for a method that returns a boolean.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,206 @@
...
+ * Responsible for file-related operations, such as reading and writing of the data file when dealing with loading
+ * and saving of task list.
+ */
+public class Storage {
+    //CHECKSTYLE.OFF: Checker
+    /** Absolute file path where the folder containing the data file is stored in. */
+    private String data_file_folder_path;
+
+    /** Absolute file path where the data file should be stored in. **/
+    private String data_file_path;
```
VietAnh1010: This naming style does not follow the convention. Can you explain the reason behind this?

src/main/java/duke/tasklist/TaskListStub.java
```diff
@@ -0,0 +1,20 @@
...
+
+/** Represents a stub for TaskList to remove dependencies during unit testing. */
+public class TaskListStub extends TaskList {
+
+    public TaskListStub() {
+    }
+
+
+    /**
+     * Return the value 3.
```
VietAnh1010: Should be `Returns` instead of `Return`.

VietAnh1010 (commented):
Overall, your code looks good. There are little to no style errors.

src/main/java/duke/command/ByeCommand.java
eugenetangkj: Okay, thanks for the suggestion!

src/main/java/duke/tasklist/TaskListStub.java
eugenetangkj: You are right!

src/main/java/duke/storage/Storage.java
eugenetangkj: Sorry, my bad. Must have been doing too much Python...

src/main/java/duke/command/EventCommand.java
```diff
@@ -0,0 +1,71 @@
...
+        //Creates task and saves it
+        Event newEventTask = new Event(taskName, startDate, endDate, DateTime.getDateTimeObject(startDate),
+                DateTime.getDateTimeObject(endDate));
+        tasks.addTask(newEventTask);
+        storage.saveTasks(tasks);
+
+        StringBuilder sb = new StringBuilder();
+        sb.append("Added task to list:\n");
+        sb.append(newEventTask.getStatusOfTaskInString() + "\n");
+        if (tasks.getSizeOfTaskList() == 1) {
```
liujiayue314: nice to see that u even took care of singular and plural nouns

src/main/java/duke/command/OnCommand.java
```diff
@@ -0,0 +1,73 @@
...
+import duke.datetime.DateTime;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.tasklist.TaskList;
+
+/**
+ * Represents an on command that is entered by the user to check what tasks are there on a specific day.
+ */
+public class OnCommand extends Command {
```
liujiayue314: Maybe can think of a better name, maybe a verb, that is more description of the action.

liujiayue314 (commented):
Overall very good work with detailed javadoc and easy to read code.

src/main/java/duke/command/OnCommand.java
eugenetangkj: Alright! Thanks for the suggestion. 👍

PR #124, opened 2023-01-23
[Gong Chen Jun] iP

0x787af25e:
# A heading

A bullet list
- bullet point 1
- bullet point 2 
- bullet point 3

A numbered list
1. entry 1
2. entry 2
3. entry 3

A fenced code block
```sql
DROP TABLE pull_requests;
```
A task list
- [x] lorem ipsum
- [ ] dolor sit amet

An emoji🥲🥲🥲 (actually 3 of them)

> A blockquote 

[A hyperlink (this is actually the kind of music I listen to, maybe I really am insane](https://www.youtube.com/watch?v=RstPyqnS9is&list=OLAK5uy_lL-Dp3s8v3NTgVaK82TY81iCn4QElR6Fc&index=9)

Inline `code`

_I am so done with this module already_


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Constructor for a Storage object
+     * @param myTaskList
+     */
+    public Storage(TaskList myTaskList) {
+        this.myTaskList = myTaskList;
+        try {
+            loadFromFile();
+        } catch (IOException ignored) {}
```
bokuanT, line 28: Would printing an error message be better than ignoring the `IOException`? Or maybe leave a comment why it is empty

src/main/java/Duke.java
```diff
@@ -1,10 +1,101 @@
+import duke.*;
```
bokuanT: Perhaps you could list all imported classes explicitly, even if it's the whole package.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+            String[] otherArgs = input.split(" ", 2);
+            String[] taskAndTime = otherArgs[1].split(" /by ", 2);
+            return new String[]{otherArgs[0], taskAndTime[0], taskAndTime[1]};
+        case EVENT:
+            otherArgs = input.split(" ", 2);
+            taskAndTime = otherArgs[1].split(" /from ", 2);
+            String[] toTime = taskAndTime[1].split(" /to ", 2);
+            return new String[]{otherArgs[0], taskAndTime[0], toTime[0], toTime[1]};
+        }
+        return null;
```
bokuanT: Just picking a nit but would you consider using this instead:
```{java}
case EVENT:
    //code
default:
    return null;
```

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Getter for whether the task is completed
+     * @return Completion status of the task
+     */
+    public boolean getCompletion() {
+        return this.isCompleted;
+    }
+    public void setCompletion() {
```
bokuanT, line 44: Perhaps naming this method as `toggleCompletion()` would be better?

bokuanT (commented):
LGTM 👍 

- Good code quality and coding standards adhered to. 

- Just need to fill in javadocs and can consider packaging into more organised packages (I used intelliJ to create new packages and refactor files to automatically package classes)

src/main/java/duke/Task.java
0x787af25e: I am mostly sure that the coding standard requires adherence to `get` + the name of the variable for boolean setter methods. Thank you for the other comments though

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,166 @@
...
+            Files.createFile(dukeFilePath);
+        } else if (!Files.exists(dukeFilePath)) {
+            Files.createFile(dukeFilePath);
+        }
+        BufferedReader dukeReader = Files.newBufferedReader(dukeFilePath);
+        String task;
+
+        while ((task = dukeReader.readLine()) != null) {
+            // process it
+            // format: E|0|project meeting|Aug 6th|2pm|4pm
```
kalarisng, line 50: i like how clean this piece of code is, thank you for including the format of the saved tasks in the file

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,105 @@
...
+                " tasks in the list!");
+    }
+
+    /**
+     * Prints a reply to standard output when the user does not enter a task.
+     * @param inputCommand What the user inputs
+     */
+    public void printReply(String inputCommand) {
+        switch (inputCommand) {
+            case "list":
```
kalarisng: minor configuration of IDE is required here, as the case clause should not be indented according to the course's java coding standard. however, the rest of the switch cases seem to follow the standard, perhaps you would like to check if this part has been reconfigured?

PR #126, opened 2023-01-23
[pangrwa] iP

pangrwa:
Alfred is your very own personal assistant,  he remembers so you can forget.
This is what makes Alfred unique:
- Simplicity
- Text-based
- Responsive and **_fast_**
- ~~Boring~~ Super fun and chill 
- Handsome butler

How can you make Alfred yours?
1. Download him from [here](https://nus-cs2103-ay2223s2.github.io/website/schedule/week4/project.html)
2. Double-click on it
3. Add your tasks for him to remember
4. Sit back and relax 🤤 

Alfred is so rich that he is willing to work for you free **FREE**!

These are some of the features that Alfred can provide:
- [x] Record your deadlines 📅 
- [x] Record any tasks 📚 
- [x] Record any of your events 🧑‍🤝‍🧑 
- [x] Save your tasks locally 💻 
- [ ] Chat with you 😉 (coming soon) 
___
If you are familiar `Java`. You can run the application using the `main` method
```java
public class Main {
    }
}
```


src/main/java/alfred/parser/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+            return new ListCommand();
+        } else if (command.equals("mark") && lineArr.length == 2) {
+            return new MarkCommand(lineArr[1]);
+        } else if (command.equals("unmark") && lineArr.length == 2) {
+            return new UnmarkCommand(lineArr[1]);
+        } else if (command.equals("delete") && lineArr.length == 2) {
+            return new DeleteCommand(lineArr[1]);
+        } else if (command.equals("find")) { // what if there's task called "find ball?"
+            return new FindCommand(lineArr[1]);
+        } else if (command.equals("list") && lineArr.length == 2) {
```
dillongoh: Would using switch cases here make this code a bit easier to read?

src/main/java/alfred/command/Command.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param storage The storage that handle the loading and writing of the tasks.
+     * @throws AlfredException The error that is thrown when the command fails to execute.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws AlfredException;
+
+    /**
+     * Checks if the command will result in the end of the program.
+     * @return True if the command closes the program, else false.
+     */
+    public abstract boolean isExit();
```
dillongoh, line 27: Since only the ExitCommand would return true for isExit(), would it be a good idea to have this method return false here so only the ExitCommand class needs to override? This would lead to less code duplication.

src/main/java/alfred/command/ListDateCommand.java
```diff
@@ -0,0 +1,48 @@
...
+
+import alfred.exceptions.AlfredException;
+import alfred.storage.Storage;
+import alfred.task.TaskList;
+import alfred.ui.Ui;
+
+/**
+ * Represents a list date command when a user wishes to list all the tasks associated to a date.
+ */
+public class ListDateCommand extends Command {
```
dillongoh, line 13: Nice! This seems like a useful command to have.

src/main/java/alfred/command/AddCommand.java
```diff
@@ -0,0 +1,101 @@
...
+                break;
+            } catch (DateTimeParseException e) {
+                throw new AlfredException("The date format should be given as dd/mm/yyyy 24hr-time\n");
+            }
+        default:
+            throw new AlfredException("I'm sorry, but I don't know what that means :<\n");
+        }
+
+        // Alfred's response to remaining tasks
+        String numTasks = tasks.getSize() == 1 ? "task" : "tasks";
```
dillongoh: Perhaps an alternate name could be used for this String variable? To me numTasks imply a number instead of a string.

dillongoh (commented):
Overall code is clear and understandable. Some nits to consider but otherwise good job!

src/main/java/alfred/parser/Parser.java
pangrwa: I agree that switch cases would make  the code a bit easier to read. Initially I was thinking of adding tasks name that include certain command names but I don't think it will be an issue anymore!

src/main/java/alfred/command/Command.java
pangrwa: Isn't Command supposed to be an abstract class?

PR #129, opened 2023-01-23
[OliviaJHL].iP

OliviaJHL:

- text-based
- easy to learn
-  ~FAST~ _SUPER_ FAST to use 🥳

All you need to do is,
1. download it from [here](https://github.com/OliviaJHL/ip/releases/tag/A-Jar).
2. run in the terminal!
3. let it manage your tasks for you 😉
4. And it is **FREE**!

Features:
- [X] Deleting tasks
- [X] Managing saving 
- [X] Managing finding tasks
- [X] Managing todo
- [X] Managing deadlines 
- [X] Managing event
- [X] Managing mark as done
- [X] Managing mark as undone

``` ruby
public class Main {
    }
}

```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+import duke.tasks.Event;
+import duke.tasks.Task;
+import duke.tasks.Todo;
+
+/**
+ * Represents commands from user input
+ */
+public class Parser {
+    private static final String Indentation = " ";
+    private static final String Horizontal = "____________________________________________________________";
```
Infrix: constant names should be in all uppercase

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+        this.ui = ui;
+    }
+
+    /**
+     * method for list, mark, unmark, delete, to do, deadline, event commands
+     * @param cmd
+     * @param  tasks
+     * @return command
+     */
+    public static Task parse(String cmd, TaskList tasks) {
```
Infrix: Perhaps you would like to create an abstract Command class for different commands to inherit from it? Then the parser could just be used to create commands, instead of executing the commands in the parser

Infrix (commented):
Overall looks good, coding standard wise I could only catch one error

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,186 @@
...
+
+import java.util.Scanner;
+
+import duke.tasks.Task;
+
+/**
+ * Represents all methods such as showLogo, showWelcome...
+ */
+public class Ui {
+    private static final String Indentation = " ";
```
jugheadjones10: Constants should be in all uppercase if I'm not wrong

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        return tasks;
+    }
+
+    /**
+     * method for updating date in saving file
+     * @param t
+     * @throws IOException
+     */
+    public void updateFile(TaskList t) throws IOException {
```
jugheadjones10: Single letter argument name (t) might make sense here because the method is short, but maybe we can guard against future confusion by just using "taskList" as the argument name instead since in the future this method might get longer.

jugheadjones10 (commented):
Looks good to merge! Some minor nits.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+                    String totime = command.substring(command.indexOf(" /to ") + 4);
+                    try {
+
+                        //Date and Time
+                        DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern((" MM/dd/yyyy HHmm"));
+                        DateTimeFormatter formatter2 = DateTimeFormatter.ofPattern((" MM/dd/yyyy HHmm"));
+                        LocalDateTime datetime1 = LocalDateTime.parse(fromtime, formatter1);
+                        LocalDateTime datetime2 = LocalDateTime.parse(totime, formatter2);
+                        //System.out.println(datetime1.format(DateTimeFormatter.ofPattern("MMM d yyyy HHmm")));
+                        //System.out.println(datetime2.format(DateTimeFormatter.ofPattern("MMM d yyyy HHmm")));
```
mandykqh: You may want to delete dead code as you can recover them from your repo if needed.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+                try {
+                   //Ui.done(words[1], tasks);
+                   str += Ui.done(words[1], tasks);
+
+                } catch (Exception e) {
+                    System.out.println(Indentation + Horizontal);
+                    System.out.println("  ☹ OOPS!!! The index number cannot be empty.");
+                    System.out.println(Indentation + Horizontal);
+
+                    str += "  ☹ OOPS!!! The index number cannot be empty. \n";
```
mandykqh: To avoid having to repeat your error messages, you can possibly throw say an `InvalidInputException` with your intended error message, and then handle it in the Duke class. The section on [A-MoreOOP](https://nus-cs2103-ay2223s2.github.io/website/se-book-adapted/projectDuke/index.html#a-moreoop) provides an example for my explanation.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+                        str += "/n";
+                    } catch (DateTimeParseException e) {
+                        //System.out.println(deadline);
+                        task = new Deadline(info, deadline, false);
+
+                        str += new Deadline(info, deadline, false);
+                        str += "/n";
+                    }
+                    tasks.add(task);
+                } catch (Exception e) {
```
mandykqh: It appears that you have many try-catch blocks here. You may consider just using a single try-catch block and then throw the necessary exceptions (eg. `InvalidInputException`, `EmptyInputException`)  that will subsequently be handled in `Duke` as mentioned in my previous comment.

mandykqh (commented):
Nice job overall! Just a few suggestions regarding error handling to reduce code duplication. 👍

PR #131, opened 2023-01-23
[pkwangjoo] iP

pkwangjoo:
## Treebot

Treebot frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is, 
1. download it from [here](https://github.com/pkwangjoo/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

****
If you are Java programer, you can use it to practice Java too. Here's the `main` method:

```Java
public class Main {
    }
}
```


src/main/java/treebot/TreeBot.java
```diff
@@ -0,0 +1,49 @@
...
+        ui.showWelcome();
+
+        while (true) {
+            try {
+                String fullCommand = ui.readCommand();
+                Command c = parser.parse(fullCommand);
+
+                if (c.equals(EXIT_TOKEN)) {
+                    break;
+                }
```
jefrai: The exit process is implemented incorrectly - the user input should be compared directly against `EXIT_TOKEN`.

```suggestion
                if (fullCommand.equals(EXIT_TOKEN)) {
                    break;
                }

                Command c = parser.parse(fullCommand);
```

src/main/java/commands/ListCommand.java
```diff
@@ -0,0 +1,13 @@
...
+
+import tasks.TaskList;
+import utils.Storage;
+import utils.Ui;
+
+public class ListCommand extends Command{
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage){
+        taskList.listTasks();
```
jefrai: `ListCommand` prints no output when the `TaskList` is empty - consider printing line dividers / some specific output so that the user knows that their command has been received and processed successfully.

src/main/java/tasks/TaskFactory.java
```diff
@@ -0,0 +1,96 @@
...
+            throw new MissingTaskDescriptionException("Task Description cannot be empty!");
+        }
+
+        String taskDescription = splitStr[1];
+        return new Todo(taskDescription);
+    }
+
+    private Task makeDeadline(String instruction) throws InvalidDateTimeFormatException {
+        String[] splitStr = instruction.split("\\s+", 2);
+        String taskDescription = splitStr[1].split(" /by ")[0];
```
jefrai: Uncaught `ArrayIndexOutOfBoundsException` is thrown when no `/by` is supplied by the user, causing the program to crash - could catch it and print an error message to the user, and then continue to run normally.

Also, not sure if the `/by` token is being located correctly - `splitStr[1]` will only contain the second word of `instruction`. Perhaps just remove the first word of `instruction` and find `/by` from there?

src/main/java/utils/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+    private String filePath;
+    private DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+
+    /**
+     * Class constructor with a given file path to the text file.
+     * @param filePath
+     */
+    public Storage(String filePath)  {
+        this.f = new File(filePath);
+        this.filePath = filePath;
```
jefrai: When the user first runs the program, the `/data` directory does not exist yet - we should create it in such a scenario to allow saving to work correctly.

```suggestion
        this.f = new File(filePath);
        f.getParentFile().mkdirs();
        this.filePath = filePath;
```

src/main/java/commands/FindCommand.java
```diff
@@ -0,0 +1,19 @@
+package commands;
+
+import exception.TreeBotException;
+import tasks.TaskList;
+import utils.Storage;
+import utils.Ui;
+
+public class FindCommand extends Command{
+    String keyword;
```
TimothyLawSongEn: Should we add an extra line below for readability?

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public Deadline(String taskDescription, LocalDateTime deadline) {
+        super(taskDescription);
+        this.deadline = deadline;
+    }
+
+
+    @Override
+    public String toStorageFormatString() {
+        return "D|" + (isDone ? "1" : "0") + "|" + this.taskDescription + "|" + this.deadline.format(DateTimeFormatter.ofPattern("d/M/yyyy HHmm"));
```
TimothyLawSongEn: Should we provide a line break for this long statement?

src/main/java/tasks/TaskFactory.java
```diff
@@ -0,0 +1,96 @@
+package tasks;
+
```
TimothyLawSongEn: Should we standardise the amount of line breaks between import statements?

src/main/java/tasks/TaskFactory.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return Task object as specified by the instruction.
+     * @throws TaskFactoryException
+     */
+    public Task make(String instruction) throws TaskFactoryException {
+        String[] splitStr = instruction.split("\\s+", 2);
+        String taskType = splitStr[0];
+
+        try {
+            switch (taskType) {
+            case "todo":
```
TimothyLawSongEn: I like how you adhered to the Switch case coding standards.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public void listTasks() {
+        for (int i = 0; i < tasks.size(); i++) {
+            System.out.println(i + 1 + "." + tasks.get(i));
+        }
+    }
+
+    /**
+     * Returns a copy of the ArrayList of Task.
+     * @return a copy of the ArrayList of Task.
```
TimothyLawSongEn: Should a line break be added before this @return statement?

src/main/java/utils/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+     * @throws TaskFactoryException
+     */
+    public Command parse(String fullCommand) throws InvalidCommandException, TaskFactoryException {
+        String[] splitStr = fullCommand.split("\\s+", 2);
+        String command = splitStr[0];
+
+        switch (command) {
+        case "list":
+            return new ListCommand();
+        case "todo":
```
TimothyLawSongEn: is this switch case required considering there is no implementation?

TimothyLawSongEn (commented):
LGTM. Just some nits to fix.

src/main/java/utils/Parser.java
```diff
@@ -0,0 +1,58 @@
+package utils;
+
+import commands.AddCommand;
+import commands.Command;
+import commands.DeleteCommand;
+import commands.ListCommand;
```
jefrai: `Parser.java` uses `commands.FindCommand` and requires an import for it.

```suggestion
import commands.FindCommand;
import commands.ListCommand;
```

src/main/java/commands/ToggleMarkTaskCommand.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param isMarkAsDone
+     */
+    public ToggleMarkTaskCommand(int index, boolean isMarkAsDone) {
+        this.index = index;
+        this.isMarkAsDone = isMarkAsDone;
+    }
+    @Override
+    public void execute(TaskList taskList, Ui u, Storage storage) throws TreeBotException {
+        if (!isMarkAsDone) {
+            taskList.unmarkTask(index);
```
jefrai: `ArrayIndexOutOfBoundsException` could be thrown here - could also catch it and inform the user.

jefrai (commented):
No major issues in how the code is structured - good work!
However, there are currently several implementation details which could be refined:

PR #137, opened 2023-01-23
[Diatbbin] ip

Diatbbin:
# DukePro

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

1.  download it from [here](https://github.com/nus-cs2103-AY2223S2/ip/pull/137).
2.  double-click it.
3.  add your tasks.
4.  let it manage your tasks for you 😉

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/duke/helper/TaskList.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Deletes the task from the tasklist
+     *
+     * @param taskNo Task number to be deleted
+     * @return task that has been deleted
+     */
+    public Task deleteTask(int taskNo) {
+        Task task = tasks.get(taskNo);
```
adam07018: Is it better to add a try catch block because there might be possible indexoutofbound error

src/main/java/duke/helper/TaskList.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Mark or unmark a task
+     *
+     * @param isDone whether the task is done
+     * @param taskId id of the task
+     */
+    public String mark(boolean isDone, String taskId) {
+        int taskNo = Integer.parseInt(taskId) - 1;
```
adam07018: Maybe can add try catch because user might input something else except number:cowboy_hat_face:

src/main/java/duke/helper/Parser.java
```diff
@@ -0,0 +1,149 @@
...
+            ArrayList<Task> output = new ArrayList<>();
+            for (Task task : taskList) {
+                if (task.toString().contains(inputs[1])) {
+                    output.add(task);
+                }
+            }
+            return Ui.filter(output);
+
+        case "bye":
+            fileSystem.updateFile(tasks);
```
adam07018: :thought_balloon:Is it necessary to update when bye command is triggered?

src/main/java/duke/helper/Parser.java
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public static LocalDateTime handleDateTime(String dateTime) throws InvalidDateTimeException {
+        try {
+            String[] period = dateTime.split("/", 3);
+            String[] yearTime = period[2].split(" ");
+            int year = Integer.parseInt(yearTime[0]);
+            int month = Integer.parseInt(period[1]);
+            int day = Integer.parseInt(period[0]);
+            int hour = Integer.parseInt(yearTime[1].substring(0, 2));
+            int min = Integer.parseInt(yearTime[1].substring(2));
```
adam07018: Suggest to use a formatter which make life much easier:innocent:

src/main/java/duke/helper/Ui.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import duke.task.Task;
+
+/**
+ * Ui class that handles all the interactions with the users
+ */
+public class Ui {
+    private final String line = "_______________________________________________________________________";
```
adam07018: Sorry I don't get the use of this:thinking:

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Constructor for creating a new Deadline task
+     *
+     * @param description desc of the task
+     * @param by task's deadline
+     * @throws InvalidDateTimeException If the dateTime input is invalid
+     */
+    public Deadline(String description, String by) throws InvalidDateTimeException {
+        super(description, false, "D");
+        this.dueDateTime = Parser.handleDateTime(by).toString();
```
adam07018: Can we do this in Parser class?

src/main/java/duke/helper/Parser.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * Parse the dateTime description into a LocalDateTime object
+     *
+     * @param dateTime dateTime description
+     * @return a LocalDateTime object
+     * @throws InvalidDateTimeException If incorrect dateTime values are given
+     */
+    public static LocalDateTime handleDateTime(String dateTime) throws InvalidDateTimeException {
```
adam07018: Should we create a time object class?

src/main/java/duke/storage/FileSystem.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Loads the tasks from the file into the TaskList
+     *
+     * @return an ArrayList that contains all the tasks from the file
+     * @throws FileNotFoundException If the file cannot be found
+     */
+    public ArrayList<Task> loadFromFile() throws FileNotFoundException, InvalidDateTimeException {
+        Scanner scanner = new Scanner(file);
```
adam07018: Nice function!

adam07018 (commented):
Review comment by adam07018

src/main/java/duke/helper/Parser.java
```diff
@@ -0,0 +1,144 @@
...
+        this.ui = ui;
+    }
+
+    /**
+     * Parse user inputs
+     *
+     * @param desc description of the task
+     * @return String representation of Task
+     * @throws DukeException If input is invalid
+     */
```
SeeuSim, line 42: It seems that a lot of logic is crammed into this method.

In the event that this method bugs out due to one of the methods, it can be difficult to find the error point and the offending switch case. 

I would suggest encapsulating the logic for each switch block in its own `Command` class, and all of them may have a `simulate` method that handles their own errors and outputs a UI regardless. This also makes it easy to define test cases for each of the commands and isolate the logic there.
Also, this would in the future make it easier to change the logic for each command, as you would only need to modify the target Command class' `simulate` method.

src/main/java/duke/helper/Ui.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+public class Ui {
+    private Scanner scanner;
+
+    public Ui() {
+        this.scanner = new Scanner(System.in);
+    }
+
+    public static String showErrorMsg(String desc) {
+        return String.format("%s commands need to be followed by an integer!\n", desc.split(" ")[0]);
```
SeeuSim: These can be abstracted into Exceptions, and all the Exceptions may be passed into a single UI method that is responsible for displaying Exceptions. You can then unit test that the exceptions are output correctly to the console.

src/main/java/duke/helper/Ui.java
```diff
@@ -0,0 +1,86 @@
...
+        return "Bye. Hope to see you again soon!\n";
+    }
+
+    /**
+     * Prints the delete message when a task is removed
+     *
+     * @param task task to be removed
+     * @param size number of remaining tasks
+     */
+    public static String showDelete(Task task, int size) {
```
SeeuSim: These can be output by a Task -> DeleteTask class that is responsible for outputting the relevant UI message to the UI. The UI then need only have a single method that parses string messages and outputs them in the desired format, or even be changed to output UI elements in the GUI.

src/main/java/duke/storage/FileSystem.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    /**
+     * Updates the file according to the tasks stored on the list
+     *
+     * @param taskList the TaskList that contains all the tasks that will be stored in the file
+     */
+    public void updateFile(TaskList taskList) {
+        try {
```
SeeuSim, line 53: This can be expensive (rewriting the entire file with each operation) if your task list grows in size. Instead, you may modify specific rows/lines in your file with each read/write operation and keep your interactions with the file *atomic*

src/main/java/duke/storage/FileSystem.java
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @return an ArrayList that contains all the tasks from the file
+     * @throws FileNotFoundException If the file cannot be found
+     */
+    public ArrayList<Task> loadFromFile() throws FileNotFoundException, InvalidDateTimeException {
+        Scanner scanner = new Scanner(file);
+        ArrayList<Task> tasks = new ArrayList<>();
+
+        while (scanner.hasNextLine()) {
+            String task = scanner.nextLine();
```
SeeuSim, line 85: this logic that is repeated in a loop may be abstracted into its own function/class so that in the event of an error, it is easy to find and fix.

SeeuSim (commented):
In general, good approaches are used in the project, but they may be further improved by:
- code splitting: Breaking the logic up into atomic chunks that handle their own errors and are easy to modify and fix.
- more test cases: Currently, events such as invalid dates being passed to Events or wrong usage of creating Events/Deadlines are not tested, and if future changes break the code, the logic will fail silently.

PR #139, opened 2023-01-23
[joyngjr] iP

joyngjr:
# **Bond**
> Marriage? Conventional happiness? I got rid of those aspirations as well as my identity, the day I became a spy – Loid Forger 

## _What is Bond?_
Bond is a to-do list management system which is
- text-based
- easy to learn
- fast to use

## _How to use it?_
1. Download files from [here](https://github.com/joyngjr/ip)
2. Run file src/main/duke/Duke.java
3. Refer to list of available user commands (type `help` to get a list of commands)
4. Add/edit tasks!

## _Features_
Experiment with it 🦙 
- [ ] Add tasks (to-do, event, deadline)
- [x] Mark/unmark tasks
- [ ] Delete tasks
- [ ] Keyword search on tasks

## _Commands Available_
```java
public static String generate() {
        return "Here is a list of commands that you may use:\n"
                + "todo <task> - creates a new task\n"
                + "deadline <task> /<yyyy-mm-dd> - creates a task with an end date\n"
                + "event <task> /<yyyy-mm-dd> /<yyyy-mm-dd> - creates a task with start and end date\n"
                + "mark/unmark <int> - change status of task\n"
                + "delete <int> - deletes a task in the list\n"
                + "find <word> - returns tasks in list which matches word\n"
                + "missions - prints out a todo list of existing tasks\n"
                + "help - prints out this list of commmands\n"
                + "bye - to close program\n";
    }
```


src/main/java/drivers/Parser.java
```diff
@@ -0,0 +1,103 @@
...
+                l.find(parts[1]);
+                return true;
+
+            case help:
+                System.out.println("B: " + Instructions.generate());
+                return true;
+
+            default:
+                System.out.println("Something has gone wrong!");
+                return false;
```
lywich: Nice use of switch statements to keep code readable.

src/main/java/support/Commands.java
```diff
@@ -0,0 +1,4 @@
+package support;
+public enum Commands {
+    bye, missions, unmark, mark, todo, deadline, event, delete, find, help
```
lywich, line 7: Have you considered capitalizing the commands? Such that a command like "Help" would be able to parse. I recommend using String.toUpperCase()/toLowerCase() method to capitalize/lowercase the words when parsing the input. 

Definitely a toss-up between rejecting commands like "heLP" or having to reject "Help". I respect the decision to only accept lowercase letters but just my two cents.

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Abstraction for data structure and operations used to maintain to do list.
+ */
+public class TaskList {
+    private ArrayList<Task> records = new ArrayList<>();
+
+    // default constructor
+    public void add(Task task) {
+        records.add(task);
+        System.out.println("B: " + task.toString() + "has been added");
```
lywich: I recommend adding a whitespace before "has been added". I notice that this might cause an issue whereby it becomes "workhas been added" unless the user themselves add a whitespace. 
```suggestion
        System.out.println("B: " + task.toString() + " has been added");
```

src/main/java/support/Storage.java
```diff
@@ -0,0 +1,41 @@
...
+public class Storage {
+    private File file;
+
+    /**
+     * Creates a directory data, and file duke.txt if it does not exist.
+     */
+    public Storage() {
+        try {
+            this.file = new File("./" + "data/duke.txt");
+            file.createNewFile();
```
lywich: Perhaps this might break when you delete the data directory? 

Deleting the data directory (that came with the zip file) results in your code being unable to recover from this deletion. Hence, the tasks are no longer stored across sessions. You might be able to observe this if the .jar file is run in 2 separate sessions. 

Just pointing this out as this seems to be a marking criteria. I recommend using Files.createDirectories() and the relative path methods like System.getProperty() and Paths.get() as a way to get around this issue.

lywich (commented):
In general, very good work. The code quality is very well done apart from some whitespaces (that is really the fault of Intellij) and one bug. I could not find any coding standard violations/unclear code. Everything is readable and design decisions are consistent. Keep up the good work!

src/main/java/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+    public String setComplete() {
+        if (this.status) {
+            return "Mission is already completed.";
+        }
+        this.status = true;
+        return "Mission Completed!\n" + this;
+    }
+
+
+    /**
```
abenx162, line 31: Should the spacing between methods be more consistent? I noticed that there's 2 lines between setComplete and setIncomplete, but none between contains and toString, for example.

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,76 @@
...
+
+        if (n == 0) {
+            return "B: There are no missions!";
+        }
+
+        for (int i = 0; i < n; i++) {
+            if (i != n - 1) {
+                s = s + (i + 1) + ". " + records.get(i).toString() + "\n";
+            } else {
+                s = s + (i + 1) + ". " + records.get(i).toString();
```
abenx162: Perhaps using a StringBuilder would be more efficient here, as we would not have to create a new String each time a new Task is appended. Although as long as the number of Tasks is small, the difference should not be significant.

src/main/java/task/Event.java
```diff
@@ -0,0 +1,20 @@
...
+        super(instruction.substring(6));
+        this.start = LocalDate.parse(start.trim());
+        this.end = LocalDate.parse(end);
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() +
+                " (from: " + this.start.getMonth() + " " + this.start.getDayOfMonth() + ", " + this.start.getYear() +
+                " to: " + this.end.getMonth() + " " + this.end.getDayOfMonth() + ", " + this.end.getYear() + ")";
```
abenx162: Perhaps the line breaks should be before the + operators, instead of after?

src/main/java/task/Event.java
```diff
@@ -0,0 +1,20 @@
+package task;
+
+import java.time.LocalDate;
+public class Event extends Task {
+    private LocalDate start;
+    private LocalDate end;
+
+    public Event(String instruction, String start, String end) {
+        super(instruction.substring(6));
```
abenx162, line 21: I think this line can be hard to understand, especially as the reader may not know that all Event instructions start with the word 'event' (and similarly for the other Task subclasses). Also, might there be issues if the instruction is just "event", and there is no 6th index? Perhaps using instruction.split(" ", 2) might be better to separate out the words.

abenx162 (commented):
overall very good! just some minor issues, other than that your code is very polished.

PR #148, opened 2023-01-23
[dunliang0513] iP

dunliang0513:


src/main/java/Duke.java
```diff
@@ -1,10 +1,49 @@
+
+import Duke.TaskList;
+import Exceptions.DukeMainExceptions;
+import Duke.Storage.Storage;
+import Duke.Ui;
+import Duke.Parser;
+
+import Exceptions.CommandNotFoundException;
+
+import java.io.IOException;
```
Tohtoroo: Your import orders are slightly off due to not being in alphabetical order, and the java.io imports should be above the duke imports!

Tohtoroo (commented):
Overall, the code was quite easy to read, just that maybe the import orders are slightly off. Possibly could switch up your Ui class to hold more functions for the interface, as it seems like most of your strings are printed from your other classes. Remember to start adding JavaDoc comments!

src/main/java/Duke/Tasks/Event.java
```diff
@@ -0,0 +1,61 @@
...
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.Period;
+import java.time.format.DateTimeParseException;
+
+public class Event extends Task {
+    protected String startingTime;
+    protected String endingTime;
+    protected LocalDate localStartingDate;
+    protected LocalDate localEndingDate;
```
antonlee59, line 14: Nice naming of fields. Makes it very clear what each field is referring to.

src/main/java/Duke/TaskList.java
```diff
@@ -0,0 +1,145 @@
...
+        System.out.println("\t--------------------------");
+    }
+
+    public void mark(int input) {
+        Task currTask = this.taskList.get(input - 1);
+        currTask.markDone();
+        System.out.println("\t--------------------------");
+        System.out.println("\tNice! I've marked this task as done:");
+        System.out.println("\t\t" + currTask.toString());
+        System.out.println("\t--------------------------");
```
antonlee59: Perhaps, all these println can be abstracted out into the Ui.java class to separate the concerns of each class better.

antonlee59 (commented):
Overall, LGTM. Maybe you can try to futher abstract out your methods especially when it comes to the Ui display as most of the printing still isn't being done in the Ui class. Good job and all the best in your upcoming tasks!

PR #157, opened 2023-01-24
[TimothyLawSongEn] iP

TimothyLawSongEn:


src/main/java/Duke/Command/AddCommand.java
```diff
@@ -0,0 +1,56 @@
...
+
+import Duke.Storage;
+import Duke.TaskList;
+import Duke.TaskType;
+import Duke.Tasks.Deadline;
+import Duke.Tasks.Event;
+import Duke.Tasks.ToDo;
+import Duke.Ui;
+
+import java.time.LocalDate;
```
ExtraShotLatte: Good use of lexical ordering for import statements

src/main/java/Duke/Command/MarkCommand.java
```diff
@@ -0,0 +1,43 @@
...
+        this.index = index;
+    }
+
+    public void execute(TaskList l, Ui ui, Storage s) throws InvalidArgumentException {
+        if (mark == 1) {
+            l.get(index).markAsDone();
+            ui.showMark(1, l.get(index));
+        } else if (mark == 0) {
+            l.get(index).markAsDone();
+            ui.showMark(1, l.get(index));
```
ExtraShotLatte: Not too sure what is the use of the `int mark` here if both does the same thing

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,168 @@
...
+//
+//
+//    public static void save(List<Task> l, String filePath) throws IOException {
+//        FileWriter fw = new FileWriter(filePath);
+//        for (Task i : l) {
+//            fw.write(i.saveFormat() + "\n");
+//        }
+//        fw.close();
+//    }
+//}
```
ExtraShotLatte: May want to consider removing this chunk of comments if there is no need for it anymore.

src/main/java/Duke/TaskList.java
```diff
@@ -0,0 +1,9 @@
+package Duke;
+
+import Duke.Tasks.Task;
+
+import java.util.ArrayList;
+
+public class TaskList extends ArrayList<Task>{
+
+}
```
ExtraShotLatte: Perharps TaskList can be used to abstract out the insertion, deletion and manipulation of the tasks?

src/test/java/Duke/Tasks/EventTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class EventTest {
+    @Test
+    public void eventToString(){
+        Event e = new Event("run", true, "sun", "mon");
+        assertEquals(e.toString(), "[E][X] run (from: sun to: mon)");
+    }
```
ExtraShotLatte: Good use of JUnit test

tasks.txt
```diff
@@ -0,0 +1 @@
+T~1~a
```
ExtraShotLatte: Not sure if adding the datafile into the repo is a good idea - ideally each local user should have their own tasklist.

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+                case "delete":
+                    return new DeleteCommand(Integer.parseInt(s[1]) - 1);
+                case "todo":
+                    return new AddCommand(TaskType.ToDo, s[1], false);
+                case "deadline":
+                    String st[] = s[1].split(" /by ", 2);
+                    return new AddCommand(TaskType.Deadline, st[0], false, LocalDate.parse(st[1]));
+                case "event":
+                    String stt[] = s[1].split(" /from ", 2);
+                    String sttt[] = stt[1].split(" /to ", 2);
```
ExtraShotLatte: Don't quite understand `sttt[]` perharps would be better to use names such as `stringToken1`, `stringToken2`?

ExtraShotLatte (commented):
LGTM! Some minor confusions on the naming of the variables but understandable nonetheless.

👍 👍 👍

src/main/java/Duke/Storage.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.Scanner;
+
+public class Storage {
+
+    private String filePath;
+    private File f;
```
jmestxr: I think the name f could be renamed to something more meaningful, e.g. file, although it's quite obvious here.
Usually use scratch variables for cases like temporary storage or loop indices.

src/main/java/Duke/Tasks/DeleteCommand.java
```diff
@@ -0,0 +1,21 @@
+package Duke.Tasks;
+
+import Duke.Command.Command;
+import Duke.Storage;
+import Duke.TaskList;
+import Duke.Ui;
+
+public class DeleteCommand extends Command {
+
+    int index;
```
jmestxr: I think the name index could be changed to something more obvious, like taskIndex to avoid confusion.

src/main/java/Duke/Command/MarkCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+import Duke.Command.Command;
+import Duke.Exceptions.InvalidArgumentException;
+import Duke.Storage;
+import Duke.TaskList;
+import Duke.Ui;
+
+public class MarkCommand extends Command {
+
+    private int mark;
```
jmestxr: I believe int mark here refers to whether the command is marking a task as done (`mark == 1`) or unmarking/marking as not done (`mark == 0`). In this case, since `mark` can only take two possible values, I think it is better if `mark` is a boolean variable. You could also add javadocs here explaining what `mark` means to avoid confusion.

jmestxr (commented):
Overall, I like how you followed the CS2103 Java code conventions most of the time. You could also start writing the javadocs for some classes/methods in the meantime for code that you think is less likely to be changed (e.g. for classes like Command).

src/main/java/Duke/Command/MarkCommand.java
TimothyLawSongEn: It was an error in my code. The second conditional statement should have been `l.get(index).markAsUndone();
            ui.showUnmark(1, l.get(index));`
Thanks for catching this error.

src/main/java/Duke/TaskList.java
TimothyLawSongEn: Can you elaborate? Because I feel that insertion and deletion etc can already be handled by ArrayList.

tasks.txt
TimothyLawSongEn: Noted. Will delete the tasks.txt file from the repo.

src/main/java/Duke/Parser.java
TimothyLawSongEn: Noted. It was bad naming on my part.

PR #161, opened 2023-01-24
[zxisatree] iP

zxisatree:
# Duke
> "In the midst of chaos, there is also opportunity.” - Sun Tzu ([source](https://www.goodreads.com/quotes/659546-in-the-midst-of-chaos-there-is-also-opportunity))

Duke is a greenfield Java project that keeps track of tasks you need to do. It:

- is text-based
- is **EXTREMELY** fast 😃
- saves your work after each command

To start using it:

1. clone the project onto your machine
2. `cd` to the root directory
3. `./gradlew run`

Features:

- [X] Managing tasks
- [X] Autosave
- [ ] GUI

It is open source, too, so if you know Java, you can edit the functionalities to do anything you want, like change the greeting message:
```java
public void showGreeting() {
    System.out.println("What are you bothering me for now?");
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,106 @@
...
+
+        Task task;
+        int taskNumber;
+
+        label:
+        while (true) {
+            try {
+                String[] parsedCommand = Parser.parseCommand(ui.readCommand());
+                String command = parsedCommand[0];
+                switch (command) {
```
Tariq-droid: very nice use of switch statement

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,106 @@
+package duke;
+
+import exceptions.DukeException;
```
Tariq-droid, line 3: I believe you need to split these imports as this is a coding standard.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,73 @@
+package duke;
+
+import exceptions.DukeException;
+
+import java.io.File;
```
Tariq-droid, line 3: split imports here as well

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+    public TaskList(ArrayList<String[]> dataFileTasks) throws DukeException {
+        this.taskList = new ArrayList<>();
+        Task toAdd;
+        for (String[] fileCommand : dataFileTasks) {
+            switch (fileCommand[0]) {
+            case "T":
+                toAdd = new ToDo(fileCommand[2]);
+                break;
+            case "D":
+                toAdd = new Deadline(fileCommand[2], Parser.parseDate(fileCommand[3], true));
```
Tariq-droid: I think "toAdd" is not a very clear name. maybe something like "newTask" is better

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Checks if the task list is empty.
+     *
+     * @return Whether the task list is empty or not.
+     */
+    public boolean isEmpty() {
+        return this.taskList.size() == 0;
+    }
+
+    private boolean indexOutOfRange(int index) {
```
Tariq-droid: because this is a boolean, I believe it should be "isIndexOutOfRange"

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,126 @@
...
+        showTextWithLines("Bye. Hope to see you again soon!");
+    }
+
+    /**
+     * Prints information about the task list after adding a task.
+     *
+     * @param task  Task that was added.
+     * @param tasks Task list that task was added to.
+     */
+    public void showAddTaskMessage(Task task, TaskList tasks) {
```
Tariq-droid: Kind of unclear method name. is it show or add?

Tariq-droid (commented):
Very nice and detailed. Just some nits in coding standards.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            System.out.println(ui.getLoadingError());
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Respond to a given input message.
```
CShuwen: A very minor point here (also for some other Javadoc): It might be good to write responds instead of respond
Quoted from coding standard:  In method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * The message to be shown before quitting the application.
+     * @return The goodbye message.
+     */
+    public String getGoodbyeMessage() {
+        return "Bye. Hope to see you again soon!";
+    }
+
+    /**
+     * Human readable information about the task list after adding a task.
```
CShuwen: Do you think it would be better to start the first sentence with a verb such as returns?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import tasks.Task;
+
+/**
+ * Handles interaction with the user and prints information about the program.
+ */
+public class Ui {
```
CShuwen, line 10: I would recommend changing the class name to UI as it is an acronym.

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,73 @@
+package duke;
+
+import java.io.IOException;
+import java.util.Collections;
+
```
CShuwen, line 5: Would recommend deleting this line in between two imports.

CShuwen (commented):
I really like how you designed the encapsulation of each method, making your code easy to read and understand. Good job!!! 👍

PR #165, opened 2023-01-24
[Glen Lim] iP

glyfy:
### Duke

- Download it
- Use it


src/main/java/command/Find.java
```diff
@@ -0,0 +1,38 @@
+package command;
+
+import tasks.TaskList;
+import tasks.Task;
+
+import java.sql.Array;
+import java.util.Arrays;
+import java.util.List;
+
+public class Find extends Command{
```
Jarrett0203: Maybe you can make the `Find` class sound more like a noun like this?
```suggestion
public class FindCommand extends Command {
```

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+    private void fileWriter(String input){
+        try {
+            FileWriter fw = new FileWriter(tasksFile, true);
+            fw.write(input);
+            fw.close();
+        } catch(IOException e){
+            System.out.println("Something went wrong " + e.getMessage());
+        }
+    }
+    private int getCount() {
```
Jarrett0203: Maybe you can change the method name to be more specific?
```suggestion
    private int getTasksFileLinesCount() {
```

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+        Task task = this.get(index);
+        task.updateState();
+        return "Ok I have marked as not done\n" + task.toString();
+    }
+
+    /**
+     * Returns size of this object
+     * @return size of this object
+     */
+    public int getSize(){
```
Jarrett0203: I am not really sure what this method is used for? Could probably just use `this.size()` directly.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Returns a message about the newly added task
+     * @param task Task object to be added
+     * @return String message about task
+     */
+    private String addTaskMsg(Task task){
+        int totalTasks = this.getSize();
+        return "Got it, I have added this task:\n" + task.toString() + "\n" +
+        "Now you have " + totalTasks + " tasks in the list.";
+    }
```
Jarrett0203: Maybe you could rename the `addTask` methods to `addTodo` for example to be more specific? Or better yet combine the `addTask` methods so code is not being repeated.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+                //load existing values into stringTaskList
+                List<String> fileContent = new ArrayList<>(Files.readAllLines(tfPath, StandardCharsets.UTF_8));
+                for (int i = 0; i < fileContent.size(); i++) {
+                    stringTaskList.add(fileContent.get(i));
+                }
+            } catch(IOException e){
+                System.out.println(e.getMessage());
+            }
+        } else{
+            try {
```
cetigerlily: i think a space is needed between catch and (IOException e) and a space after else - just to follow the coding standard

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,172 @@
+package tasks;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
```
cetigerlily, line 9: you could write down the exact java.util imports since wildcard imports aren't recommended to be used

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+                        break;
+                    }
+                }
+            } catch(IOException e){
+                throw new DukeException("Failed to load tasks from storage");
+            }
+        }
+    }
+
+    /**
```
cetigerlily, line 69: great to see all the javadocs you have added!

cetigerlily (commented):
keep up the good work! i think just some small editing to adhere to the coding standard is needed

PR #174, opened 2023-01-24
[kalarisng] iP

kalarisng:
# DUKE
> "Everything begins with an idea." - Earl Nightingale

- text-based
* easy to learn
+ ~~FAST~~ SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)


src/main/java/duke/EmptyDescription.java
```diff
@@ -0,0 +1,15 @@
+package duke;
+
+/**
+ * Represents an Exception due to absence of description following task.
+ */
+public class EmptyDescription extends Exception {
```
gitsac, line 6: I like that you abstracted empty user inputs into its own class!

src/main/java/duke/WrongTask.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+/**
+ * Represents an Exception due to user inputting a wrong task keyword.
+ */
+public class WrongTask extends Exception {
```
gitsac, line 6: Would it be better to centralize a DukeExceptions class which can then be inherited by the various separate Exception classes?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,38 @@
...
+     * Creates a new Deadline task.
+     * @param description Description of the task.
+     * @return Deadline task.
+     * @throws DateTimeParseException If the date is in the wrong format.
+     */
+    public static Deadline createDeadline(String description) throws DateTimeParseException {
+        String[] split = description.split("/by ");
+        LocalDate d1 = LocalDate.parse(split[1]);
+        return new Deadline(split[0], d1.format(DateTimeFormatter.ofPattern("MMM d yyyy")));
+    }
```
gitsac: Would it be better to make the Deadline constructor private if there is going to be a factory method to create Deadline tasks?

gitsac (approved):
I like that you have abstracted the various exceptions for different cases into their own classes, very neat, but you might want to consider whether a centralized class for Exceptions pertaining to Duke whereby the other classes inherit from would bring it up a notch.
Overall, the rest of your iP looks good and well-documented in terms of your JavaDocs!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Main class.
+ */
+public class Duke {
+
+    private final TaskList taskList;
+    private final Ui ui;
+    private final Storage storage;
+    private static final String filePath = "data/duke.txt";
```
wpx12011, line 11: Should the constant filePath be named as 'FILEPATH'?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,38 @@
+package duke;
```
wpx12011, line 1: Good job done for putting every class in a package!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,190 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
```
wpx12011: Good job done for having specific import statement.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Checks if a task has an empty description.
+     *
+     * @param checkString User input to be checked.
+     * @return True or false depending on whether description of task is empty.
+     * @throws EmptyDescription If the description of task is empty.
+     */
+    public static boolean checkEmptyDescription(String[] checkString) throws EmptyDescription {
```
wpx12011, line 264: Perhaps the boolean method could be named as 'isEmpty'?
I think the method will always return true, and the usage of it is to throw an exception when the string description is empty. May be can set the method to be void so there won't be a need to return anything?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,45 @@
...
+     * Marks the task as done.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Unmark the task as done.
+     */
+    public void unMarkAsDone() {
```
wpx12011, line 39: I could understand that the naming unMarkAsDone somewhat gives it a better contrast with markAsDone haha, but perhaps need to change to unmarkAsDone() since unmark is a word itself.

wpx12011 (approved):
Nice OOP code and I could follow the structures clearly, great job!

PR #178, opened 2023-01-24
[KevinEyo1] iP

KevinEyo1:

- text-based
- easy to learn
- ~~FAST~~ **_SUPER_ FAST** to use
All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```java
`public class Main {
    }
}
```


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,85 @@
...
+            return new AddEventCommand(temp[0], temp2[0], temp2[1]);
+        } else if (cmd.equals("delete")) {
+            if (len == 1) {
+                throw new DukeException("Unable to delete.");
+            }
+            return new DeleteCommand(details);
+        } else if (cmd.equals("find")) {
+            return new FindCommand(details);
+        } else {
+            throw new DukeException("I do not quite understand that.");
```
linustws: Perhaps can consider using switch case instead of if-else? I heard that it's faster, especially when the list gets longer.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,35 @@
...
+    public DeleteCommand(String details) {
+        this.index = Integer.valueOf(details);
+    }
+    /**
+     * Executes the command given by the user.
+     *
+     * @param tasks   to be modified
+     * @param ui      to display changes
+     * @param storage to interact with as necessary
+     * @throws DukeException
```
linustws: Should the parameter description start with caps and have a fullstop at the end? Also, should they be indented at the same level or be just right after the parameter name? Perhaps can add why the DukeException might be thrown at the back of @throws, according to the Java coding standard. I noticed this same issue in several other places too.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,34 @@
...
+     * Executes the command given by the user.
+     *
+     * @param tasks   to be modified
+     * @param ui      to display changes
+     * @param storage to interact with as necessary
+     * @throws DukeException
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        ui.showMatchingTasks(tasks.filterTasks(this.findString));
```
linustws: I like how you separated the ui and the tasklist such that there's a command for the ui to print each respective action that the tasklist executed, instead of letting the tasklist print it. This embodies the concept of abstraction. 👍

linustws (commented):
LGTM, almost. Just a few nits to fix. 😃

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Displays divider line.
+     */
+    public void showLine() {
+        System.out.println(DIVIDER);
+    }
+
+}
```
igezt: Good work on abstracting out the many methods for your Ui!

src/test/java/duke/data/TaskListTest.java
```diff
@@ -0,0 +1,29 @@
+package duke.data;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import duke.data.task.Todo;
+
+public class TaskListTest {
+    @Test
+    public void testAddTask() {
```
igezt: I think you can add more details to the test method using the following convention:
whatIsBeingTested_descriptionOfTestInputs_expectedOutcome

src/test/java/duke/data/TaskListTest.java
```diff
@@ -0,0 +1,29 @@
...
+        TaskList tasks = new TaskList();
+        tasks.add(new Todo("Do homework"));
+        assertEquals(1, tasks.size());
+        assertEquals("T | 0 | Do homework", tasks.get(0).storageStr());
+    }
+
+    @Test
+    public void testDeleteTask() {
```
igezt: Same as above! Adding more details will help you remember what the testing method is testing

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,93 @@
...
+        try {
+            Scanner s = new Scanner(f); // create a Scanner using the File as the source
+            ArrayList<Task> tasks = new ArrayList<Task>();
+            while (s.hasNext()) {
+                // convert text to duke.data.task.Task
+                String[] input = s.nextLine().split(" \\| ");
+                String type = input[0];
+                String mark = input[1];
+                Task x = null;
+                if (type.equals("T")) {
```
igezt: Can use a switch-case statement here!

igezt (commented):
Overall good! Use switch-case instead of if-else. And name your test-cases a little better!

src/main/java/duke/command/DeleteCommand.java
KevinEyo1: The coding standard seems to state as such.

Good point pointing out about the @throws.

src/main/java/duke/command/FindCommand.java
KevinEyo1: Thanks!

> I like how you separated the ui and the tasklist such that there's a command for the ui to print each respective action that the tasklist executed, instead of letting the tasklist print it. This embodies the concept of abstraction. 👍

src/main/java/duke/parser/Parser.java
KevinEyo1: You are right!

> Perhaps can consider using switch case instead of if-else? I heard that it's faster, especially when the list gets longer.

src/main/java/duke/ui/Ui.java
KevinEyo1: Thanks!

src/test/java/duke/data/TaskListTest.java
KevinEyo1: Yes, I will take note of this.

src/test/java/duke/data/TaskListTest.java
KevinEyo1: Could you could provide an example?

src/main/java/duke/storage/Storage.java
KevinEyo1: Yes, you are right!

src/test/java/duke/data/TaskListTest.java
igezt: I meant using whatIsBeingTested_descriptionOfTestInputs_expectedOutcome for the naming

PR #179, opened 2023-01-24
[Tan Yu Fei] iP

Lanreath:
# DukePro
> "Duke is the greatest piece of software ever written." - Lanreath
DukePro frees your mind of having to remember things. Just do it.
- text-based
- easy to learn
- ~~FAST~~ _REAL_ FAST
All you need to do is:
1. download it from [here](https://nus-cs2103-ay2223s2.github.io/website/schedule/week4/project.html)
2. double-click it.
3. add your tasks.
4. let it remember tasks for you :grimacing:
And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---
If you a Java lover, you can use it practice Java too. (_but why would you?_) Here's the `main` method:
```
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    private File file;
+
+    /**
+     * Constructor for Storage class.
+     * 
+     * @param filePath the path of the file to save and load tasks from.
+     * @throws DukeException
+     */
+    public Storage(String filePath) throws DukeException{
```
Varstak: Is there a missing whitespace between `DukeException` and `{`

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * Parses a date string into a LocalDate object.
+     * 
+     * @param date the date string to parse.
+     * @return the LocalDate object.
+     * @throws DukeException 
+     */
+    public static LocalDate parseDate(String date) throws DukeException {
```
Varstak: Would the method name `stringToLocalDate` instead of `parseDate` be more clear?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,116 @@
...
+                throw new DukeException("Please provide a start time for this event.");
+            }
+            String[] tokens2 = tokens[1].split(" /to ", 2);
+            if (tokens2.length == 1) {
+                throw new DukeException("Please provide an end time for this event.");
+            }
+            return new CreateEvent(tokens[0], parseDate(tokens2[0]), parseDate(tokens2[1]));
+        } else {
+            throw new DukeException("I'm sorry, but I don't know what that means :-(");
+        }
```
Varstak: Could using `switch` statements help with additions to features in the future?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,140 @@
...
+    public void showLine() {
+        System.out.println("----------------------------------------");
+    }
+
+    /**
+     * Display goodbye message.
+     */
+    public void showBye() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
```
Varstak: Any reason you implemented these methods rather than saving the strings as constants?

Varstak (commented):
Not too many coding standard issues. Keep it up !!! :smile::smile::smile:

src/test/java/duke/StorageTest.java
```diff
@@ -0,0 +1,72 @@
...
+            storage.save(list);
+            Scanner s = new Scanner(testFile);
+            assertEquals("T | 0 | test", s.nextLine());
+            assertEquals("D | 1 | test | 2020-01-01", s.nextLine());
+            assertEquals("E | 0 | test | 2020-01-01 | 2020-01-02", s.nextLine());
+            s.close();
+        } catch (IOException e) {
+            fail("IOException thrown");
+        } catch (DukeException e) {
+            fail("DukeException thrown");
```
Varstak: Could making this string a constant in the parent `DukeException` class make the code easier to read and reduce bugs? I noticed you repeated the same string multiple times in this class.

src/main/java/duke/Parser.java
Lanreath: Using an action verb for a method name seems more suitable.

src/main/java/duke/Storage.java
Lanreath: Good catch!

src/main/java/duke/Parser.java
Lanreath: That is a great suggestion, will include it in the next version 😄

src/test/java/duke/StorageTest.java
Lanreath: I wanted to leave room for future custom messages, but that could work if the messages end up staying the same.

immanuelhume:
Hi I came across your project and found it really cool since it is written in Java, it is 100% memory safe and blazingly fast :rocket: reliable software. so I would like to license DukePro to use it for my new startup, we want to switch from notion written in slow JavaScript to DukePro, can you please contact me and we can work out an arrangement thank you, regards.

PR #182, opened 2023-01-24
[Oliverloo0909] iP

Oliverloo0909:
### DukeBot

> "A goal without a plan is just a wish." Antoine de Saint-Exupéry [source](https://www.quotespedia.org/authors/a/antoine-de-saint-exupery/a-goal-without-a-plan-is-just-a-wish-antoine-de-saint-exupery/)

DukeBot makes planning so much easier, so you can finally reach your goals. It is: 

- command-line friendly 
- intuitive 
- simple
- ~beautiful~  **ULTRA SEXY** user interface

_You're only one step away from making planning so much more fun and easy!_ 

All you need to do is download it from [here](https://github.com/Oliverloo0909/ip/releases/tag/A-Jar).

1. wake up duke
2. tell it your tasks 
3. That's it! 

You'll never worry about missing any tasks 😉
it is **FREE**!

Features:
- [ ]  Managing tasks
- [ ]  Managing deadlines 
- [ ] Managing Events 

How easy is it to use Duke? If you 're a Java programmer, it's easier than writing this code out!

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,224 @@
...
-        System.out.println("Hello from\n" + logo);
+    /**
+     * Main method to initiliase duke chatbot
+     * @param args
+     */
+        public static void main (String[]args) {
+            Tasklist list = new Tasklist();
+            try {
+                Storage.loadFile(list);
+            } catch(Exception e) {
```
jinbesan: A catch for an "everything" Exception, maybe you could instead make Storage.loadFile throw a more specific DukeException instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,224 @@
...
+
+                    //event
+                    else if (Parser.is_Event(input)) {
+                        Duke.executeEvent(input, list);
+                    }
+
+                    //find
+                    else if (Parser.is_Find(input)) {
+                        Duke.executeFind(input, list);
+                    }
```
jinbesan: It's good that the boolean method names starts with an is, but I think the method names could be corrected to use camelCase to adhere to the coding style, for example isList(), isBye() etc.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,121 @@
...
+ * Handles the user interface and overall appearance of the duke program
+ * @author oliverloo
+ * @version 1.0
+ *
+ */
+public class Ui {
+    static String line = "      _____________________________________________________________________";
+
+    /**
+     * Greet user interface: Duke greets the user
```
jinbesan: Javadocs summary should start with the form of "Greets" instead

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,60 @@
...
+        assertEquals(true, duke.Storage.strToBool("1"));
+    }
+
+    @Test
+    public void strToBoolTest2(){
+        assertEquals(false, duke.Storage.strToBool("0"));
+    }
+
+    @Test
+    void givenSystemOutRedirection_whenInvokePrintln_thenOutputCaptorSuccess() {
```
jinbesan: good test method naming!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,51 @@
...
+        this.isMarked = isMarked;
+    }
+
+    public boolean isMarked() {
+        return isMarked;
+    }
+
+    public String get_content () {
+        return content;
+    }
```
benjamin-wee: Perhaps you can used camel case naming convention here as per the coding standard!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void unmark() {
+        isMarked = false;
+        Ui.displayUnmark(this);
+    }
+
+    public abstract String addDivider();
+
+    public boolean is_Marked() {
```
benjamin-wee: Perhaps omit the use of an underscore here to adhere to the coding standard!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,215 @@
...
+        String todo = input.split(" ", 2)[1];
+        return todo;
+    }
+
+    /**
+     * Checks if input is todo task command
+     * @param input
+     * @return true if input is todo task , else false
+     */
+    public static boolean is_toDo(String input) {
```
benjamin-wee: Perhaps omit the underscore here and use camelCase here to adhere to the coding standard!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,80 @@
...
+     * @param str
+     * @return boolean value, true if 1, false if 0
+     */
+    public static boolean strToBool(String str) {
+        if (str == "1") {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
benjamin-wee: Perhaps you could name the method here differently ie isTrue to make the method sound like a boolean as per the coding standard!

benjamin-wee (commented):
In general the code is well written, however perhaps some minor edits can be done for naming the variables/methods to adhere to the coding standards better! Keep up the good work!

PR #191, opened 2023-01-24
[Diamondroxxx] iP

EepyKitty:
## CatBot

> This is a fancy quote that exists for no reason 

CatBot has a few key properties
* it's **fast**
* it's **efficient**
* it's a **cat**!
* ~~it also does your homework~~ _(unfortunately this is a joke)_

To use it, follow these easy steps:

1. Download the latest release from [here](https://github.com/Diamondroxxx/ip/releases/tag/A-Jar).
2. Navigate to the folder using a terminal
3. Run `java -jar catbot.jar`
4. Enjoy! 🎉

Current commands include

- [x] todo
- [x] deadline
- [x] event
- [x] list
- [x] mark
- [ ] unmark
- [x] echo
- [ ] find

-----
### For Java programmers

If you happen to be a Java programmer and want to integrate `CatBot` into your program, here is how to make a `CatBot` instance.

```java
    CatBot bot = new CatBot("save/file/location");
}
```


src/main/java/catbot/CatBotException.java
```diff
@@ -0,0 +1,15 @@
...
+public class CatBotException extends Exception {
+
+    /**
+     * Initialises a CatBotException.
+     * @param message is the message to show the user
+     */
+    public CatBotException(String message) {
+        super(message);
+    }
+}
```
swxk19: CatBotException can be split into more specific exceptions, so a specific exception can be thrown instead of general error. e.g: ListIndexOutOfRangeException if the user tries to invoke a command outside of the list's range

src/main/java/catbot/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+    public static Command parse(String command) throws CatBotException {
+        String[] cmd = command.split(" ", 2);
+        String[] temp;
+        switch (cmd[0].toLowerCase(Locale.ROOT)) {
+        case "todo":
+            try {
+                return new AddCommand(cmd[1].strip());
+            } catch (ArrayIndexOutOfBoundsException e) {
+                throw new CatBotException("That's the wrong format!");
+            }
```
swxk19: It might be better to move the error handling to within the individual method commands
So that in future, if you decide to do it, you can have more specific error handling for each kind of task created. E.g: "That's the wrong format for an Event! Event should have the format: event <task_name> /from ....."

src/main/java/catbot/tasklist/EventTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private String formatDate(LocalDateTime date) {
+        return date.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM, FormatStyle.SHORT));
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString()
+                + " (" + formatDate(startsAt)
+                + " – " + formatDate(endsAt) + ")";
```
swxk19, line 28: Might be cleaner to use String.format() :)

swxk19 (commented):
Great Job!

src/main/java/catbot/commands/Command.java
```diff
@@ -0,0 +1,28 @@
...
+        return isExit;
+    }
+
+    protected void setExit() {
+        isExit = true;
+    }
+
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws CatBotException;
+
+    public void loadCommand(ArrayList<Task> tasks) throws CatBotException {}
```
Bisceto, line 27: I think you can declare loadCommand as an abstract method.

src/main/java/catbot/commands/DeleteCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import catbot.storage.Storage;
+import catbot.tasklist.Task;
+import catbot.tasklist.TaskList;
+import catbot.ui.Ui;
+
+/**
+ * Handles deleting a task from the task list.
+ */
+public class DeleteCommand extends Command {
+    private final int index;
```
Bisceto, line 15: Since this attribute is final, it may be better to make 'index' all uppercase

src/main/java/catbot/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+public class Parser {
+
+    /**
+     * Parses an input string to figure out the command.
+     * @param command is the user input for a command.
+     * @return a {@code Command} object for the user command.
+     * @throws CatBotException if the user input is malformed.
+     */
+    public static Command parse(String command) throws CatBotException {
+        String[] cmd = command.split(" ", 2);
```
Bisceto: Variable names for collections should be in plural (and meaningful as well)

src/main/java/catbot/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+                throw new CatBotException("That's the wrong format!");
+            }
+
+        case "bye":
+            return new SaveCommand();
+
+        default:
+            throw new CatBotException("I don't know what you mean.");
+        }
+    }
```
Bisceto, line 146: Good use of switch statements in parsing the command! If you can handle the exceptions internally with each command, it will definitely make the code look cleaner

src/main/java/catbot/storage/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+        try {
+            Scanner scanner = new Scanner(saveFile);
+
+            while (scanner.hasNextLine()) {
+                Parser.parse(scanner.nextLine()).loadCommand(tasks);
+            }
+
+            scanner.close();
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
Bisceto, line 62: You can throw a CatBotException to make the error handling more meaningful.

src/main/java/catbot/tasklist/Task.java
```diff
@@ -0,0 +1,56 @@
+package catbot.tasklist;
+
+import catbot.ui.ConsoleColors;
+
+/**
+ * The parent class of all tasks.
+ */
+public abstract class Task {
+    protected boolean isDone;
+    final String description;
```
Bisceto, line 11: You should make 'description' either protected or private. By default, Java leaves unspecified variables as public and this can violate information hiding

Bisceto (commented):
LGTM! Just a few nits to fix.

PR #192, opened 2023-01-24
[Jerome Neo] iP

jerome-neo:
<h1>Duke+</h1>

> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes/)

### Duke+ helps you organise your life, perfect for rememebering:
<li> Events
<li> Deadlines
<li> Todos

~~CROSS~~ out your list _anywhere_ and _anytime_.

Follow these steps:
1. download [here](https://github.com/jerome-neo/ip/releases/tag/A-Jar).
2. double-click
3. key in your commands
4. organise your life!

Most importantly, it is **FREE**!

Road map:
- [x] Basic features
- [ ] Setting reminders
- [ ] UI improvements

Are you a developer?
Here's how you can help improve the app!
```
public class Main {
    }
}
```


src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,73 @@
...
+                display.line();
+            } else if (input.equals("bye")) {
+                display.farewell();
+                break;
+            } else if (input.equals("list")) {
+                tasks.printTaskList();
+                display.line();
+            } else if (input.startsWith("mark")) {
+                int taskNumber = logic.indexToMark(input);
+                tasks.taskMarkedAtIndex(--taskNumber);
```
leeyi45: I personally don't like using the prefix increment thing, because it's usually terribly confusing (increment then return? return then increment?)

src/main/java/duke/store/Storage.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    /**
+     * Returns a string object of how the task object will be saved in the save file.
+     *
+     * @param task is the task to be converted into a String to be saved.
+     * @return string that is to be saved.
+     */
+    private static String readerFriendly(Task task) {
```
leeyi45, line 101: Methods should use verbs. Maybe `getReaderFriendly()`?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.time.LocalDateTime;
+
+/**
+ *      File name: Deadline.java
+ *      @author: Jerome Neo
+ *      Description: Deadline class that inherits from the Task class.
+ */
+public class Deadline extends Task {
+    protected String by;
```
leeyi45, line 11: (personally idm this) but there are people who think that fields (especially non final ones) should always be private

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Returns a LocalDateTime object that can be stored.
+     * This method accepts Strings with the following format dd/MM/yyyy HHmm
+     * where the time is in 24 hours.
+     * @param stringDate
+     * @return the LocalDateTime representation of the string.
+     */
+    public static LocalDateTime convertDateTime(String stringDate) {
```
leeyi45, line 110: I think you should follow the format of your similarly named method `dateTimeToString` and name this `stringToDateTime`

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+     * Returns a duke.task.TaskList of tasks matching the string description.
+     *
+     * @param description of a task.
+     */
+    public void matchDescription(String description) {
+        System.out.println("Here are the tasks matched in your list: ");
+        Integer j = 0;
+        for (int i = 0; i < tasks.size(); i++) {
+            Task task = this.getTaskAtIndex(i);
+            if (task.getDescription().equals(description)) {
```
leeyi45: Consider using `contains`, or some kind of locale based comparisons

src/data/duke.txt
```diff
@@ -0,0 +1,2 @@
+Todo: run     ๑(◕‿◕)๑ COMPLETED! ๑(◕‿◕)๑    
+Deadline: interview internship by 01/01/2023 1300     (｡-_-｡ ) INCOMPLETE ( ｡-_-｡)    git
```
vsamarth: Perhaps removing this from the repository by adding it to .gitignore would be suitable as it is something specific to each user.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ *      File name: Task.java
+ *      @author: Jerome Neo
+ *      Description: Task class that Todo, Event and Deadline inherits from.
+ */
+public class Task {
```
vsamarth, line 11: Consider marking this class abstract as there are no instantiations of this class and its an abstract entity that other classes inherit from.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,54 @@
+package duke.ui;
+
+/**
+ *      File name: duke.ui.Ui.java
+ *      @author: Jerome Neo
+ *      Description: duke.ui.Ui class deals with interaction with the user.
+ */
+public class Ui {
+    private final String LINE = "____________________________________________________________";
+    private final String LOGO = " ____        _        \n"
```
vsamarth: These can be marked as static as they have fixed value.

src/main/java/duke/main/Duke.java
vsamarth: I agree with this. Decrementing it first and then using this would be much more clearer.

vsamarth (approved):
The project overall is very well structured and the focus on details is excellent. 

However, the only thing I noticed was that the abstraction between tasks and the UI are not maintained properly (there are print statements in the task classes).
This would make changing the User Interface more tricky as you would have change your task classes as well.

PR #194, opened 2023-01-24
[WindyCall] iP

WindyCall:
WindyCall frees your mind of having to remember things you need to do. It's,
- Text-based
* Command-base
+ ~~CONVENIENT~~ SUPER CONVENIENT to use

All you need to do is,
1. download it from [here](https://github.com/WindyCall/ip/releases/tag/A-Gradle)
2. drag it to the folder you want to use
3. type: **java -jar windycall-1.0.jar** in command line to get started
4. let it manage your tasks for you :smile:

One thing to notice: It is **FREE**

Features:
- [x] Managing tasks
- [x] Managing deadlines/events
- [ ] Reminders (coming soon) 

```java
    WindyCall chatBox = new WindyCall();
    chatBox.run();
}
```


src/main/java/windycall/Deadline.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class Deadline extends Task{
+
+    private LocalDate deadline;
+
+
+    public Deadline(String description, boolean status, String deadline) {
+        super(description, status);
+        this.deadline = processDateTime(deadline);
+    }
```
yyj-02: Can we move the `processDateTime` to `Parser`? In my opinion, it would be better to pass `deadline` in as `LocalDate` and let `Parser` handles the `processDateTime` as this will be more fitting with the Single Responsibility Principle.

src/main/java/windycall/Deadline.java
```diff
@@ -0,0 +1,65 @@
...
+        return dateTime;
+    }
+
+    private String getDeadlineFormat() {
+        return deadline.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.LONG));
+    }
+
+    private String getDeadline() {
+        return " (by: " + getDeadlineFormat() + ")";
+    }
```
yyj-02: Can we change these 2 method names to make them more distinguishable while explaining the respective functionality?

src/main/java/windycall/Event.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+
+    @Override
+    public String getFileFormat() {
+        return "E | " + getStatusIcon() + " | " + description + " | " + from + " | " + to + "\n";
+    }
+    private String getDeadlineFormat(LocalDate date) {
+        return date.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.LONG));
+    }
```
yyj-02: There is a `getDeadlineFormat` method which I think might be confused with the methods in `Deadline` class. Can we change the method name?

src/main/java/windycall/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return num;
+    }
+
+    public int getDeleteIndex(String[] parts) {
+        if (parts.length == 1) {
+            System.out.println("     You should input a number to delete a task");
+            return -1;
+        }
+        int num = -1;
```
yyj-02: Can we use a more specific name, e.g. `index`? Similarly in other variables relating to the index.

src/main/java/windycall/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+            //break;
+        case "delete":
+            return OperationType.DELETE;
+            //break;
+        case "find":
+            return OperationType.FIND;
+            //break;
+        default:
+            return OperationType.ADDTASK;
+            //break;
```
yyj-02: You can consider using `Enum.valueOf()` method and `String.toUpper()` to achieve this in shorter lines of code. 😊

src/main/java/windycall/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Loads data inside data file to current tasks.
+     * Handles non-existence of files or related directory
+     *
+     * @param tasks an ArrayList that will be loaded into
+     */
+    public void handleLoad(List<Task> tasks) {
+        //check if "./data" directory exist
+        File file = new File("./data");
+        File task_file = new File("./data/WindyCall.txt");;
```
yyj-02: Can we change this to camelCase?

src/main/java/windycall/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+
+    }
+
+    /**
+     * Handles tasks change by overwriting contents in tasks
+     * back to data file so that change is saved each time tasks get modified
+     *
+     * @param tasks tasks that has been changed to do users' operation on tasks
+     */
+    public void handleTaskChange(List<Task> tasks) {
```
yyj-02: `handleTaskChange` does not really encapsulate the functionality of this method. Should we use `writeTasksToFile` instead?

src/main/java/windycall/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static void space() {
+        System.out.print("     ");
+    }
+
+    public static void line() {
+        space();
+        System.out.println("--------------------------------------------------");
+    }
```
yyj-02: For method names, it is best to use an action verb. Can we change it to `printSpace` and `printLine` respectively?

src/main/java/windycall/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+        System.out.println("Here are all matching tasks in your list:");
+        for (int i = 0; i < tasks.size(); i++) {
+            if (tasks.get(i).matchFilterWord(filterWord)) {
+                Ui.space();
+                System.out.println((i + 1) + "." + tasks.get(i));
+            }
+        }
+    }
+
+    public void greeting() {
```
yyj-02: Similarly for this

src/main/java/windycall/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+        line();
+        space();
+        System.out.println("Hello! I'm WindyCall");
+        space();
+        System.out.println("How can I help you?");
+        line();
+    }
+
+    public void byeWords() {
```
yyj-02: And this too

src/test/java/windycall/DeadlineTest.java
```diff
@@ -0,0 +1,25 @@
+package windycall;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+    @Test
+    public void standardDateFormatTest() {
```
yyj-02, line 14: I think the usual convention for testing method names should start with `should...` to say its expected behaviour, i.e. `shouldParseFormattedDate`. Do you want to follow that convention?

src/main/java/windycall/Task.java
```diff
@@ -0,0 +1,52 @@
...
+ * Serves as an abstraction of different types of tasks
+ */
+
+public abstract class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description, boolean status) {
+        this.description = description;
+        this.isDone = status;
```
yyj-02: Can we change `status` to `isDone` also?

yyj-02 (commented):
Well done! 👍👍 I think your variable namings are generally very well thought out and self-explanatory. If I have to nitpick, however, I would suggest that you can rename the testing method names to start with `should...`, stating the expected behaviour. But overall very good effort! 🔥

src/main/java/windycall/Deadline.java
WindyCall: Nice suggestion and it makes much sense!!

src/main/java/windycall/Deadline.java
WindyCall: Yes, I think I haven't thought deep when naming these two methods. Thanks

src/main/java/windycall/Parser.java
WindyCall: Sure, I will have a try.

src/main/java/windycall/Task.java
WindyCall: I've looked up all your suggestions and they are really helpful for me. Thanks for your kind and careful reminder!!!

Btw, have we met before, why I felt like you are quite familiar hahah?

src/main/java/windycall/Ui.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+public class Ui {
+
+    public Ui() {
+
+    }
+
+    public static void space() {
+        System.out.print("     ");
+    }
```
PeanutButters93: I notice a lot of ui.space() followed by System.out.println().
Perhaps consider abstracting further and have a ui.println() ?
It would also make it more clear that space is writing something to the user.

src/test/java/windycall/DeadlineTest.java
PeanutButters93: I think the method name is fine.
According to the [Coding Standard](https://se-education.org/guides/conventions/java/intermediate.html), the test function can be featureUnderTest_testScenario_expectedBehavior(), where the testScenario and expectedBehaviour are optional.

src/main/java/windycall/WindyCall.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Serves as a trigger of the whole WindyCall chatBox application.
+ * Stores some important helper classes instance.
+ */
+public class WindyCall {
+
+      private List<Task> tasks;
+      private Storage storage;
+      private Ui ui;
```
PeanutButters93: Having the Ui object be called ui and frequently calling both could be confusing. Might I suggest renaming the ui to uiObject? If you choose to do so, there is a feature in Intelij to automatically edit all instances of a phrase (shift+F6).

src/main/java/windycall/OperationHandler.java
```diff
@@ -0,0 +1,7 @@
+package windycall;
+
+public abstract class OperationHandler {
+
+    public abstract void handle(String userCommand);
```
PeanutButters93: I see there are static handle functions and an instance handle functions. Maybe consider renaming or removing one? I don't believe this (static) is used. I believe you may have tried to override this static function. I also tried that and realised it was not possible.

PeanutButters93 (commented):
Looks very good.
Some minor changes suggested.

src/main/java/windycall/WindyCall.java
WindyCall: Sure, it makes sense. Thanks for that.

src/main/java/windycall/OperationHandler.java
WindyCall: Yes, I do have this issue when coding. So far I just use static to make it functionable, and I think no need to do the renaming because I can overload this handle methods specifically in different classes while maintaining ```public void handle(String)```.
Anyways, thanks for sharing your own experience, I'll have a try!!

PR #197, opened 2023-01-24
[Statspadders] iP

Statspadders:
# Welcome to Duke
>The famous Ryochi once said that in your life, you need to have Duke to store all your tasks.

You can download it [here](https://github.com/Statspadders/ip)

Benefits of Using Duke
+ Free
+ Pay to win
+ Fastest program

1. Best
2. LOL
3. Pro

```ruby
require 'redcarpet'
markdown = Redcarpet.new("Hello World!")
puts markdown.to_html
```

- [X] Hello world
- [ ] Testing
- [ ]  What

@octocat :+1: This PR looks great - it's ready to merge! :shipit:

`Hello world` `Java` `Donezo`

_testing_


src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
+import duke.*;
```
linustws: Perhaps import each class explicitly, as per the Java coding standard. I noticed the same issue in several other places too.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        return list;
+    }
+
+    /**
+     * Write a new line to the file
+     * @param t = task
+     */
+    public void write(Task t){
```
linustws, line 69: Perhaps 'save' would be a better function name than 'write'? because it goes with 'load'. Or maybe changing 'load' to 'read', such that it matches better.

src/main/java/duke/command/CommandNotFound.java
```diff
@@ -0,0 +1,27 @@
...
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
+/**
+ * CommandNotFound - when user enters a Command and Duke
+ *         does not understand the command.
+ */
+public class CommandNotFound extends Command {
```
linustws: Perhaps consider changing the name of this class to be InvalidCommand or UnknownCommand, to be consistent with the naming of the rest of the command classes.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+    public void customMessage(Task value) {
+        System.out.println(value);
+    }
+
+    public String readCommand() throws IOException {
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+        String value = br.readLine();
+        return value;
+    }
+    public void showMarkSucess(Task value) {
```
linustws: Consider rectifying the spelling/gramatical error here to showMarkSuccessful or maybe just change it to showMarked, makes it less lengthy yet concise. Also, isn't there supposed to be an unmark command too?

linustws (commented):
LGTM, almost. Just a few nits to fix. 👍

Statspadders:
> LGTM, almost. Just a few nits to fix. 👍

Thank you for your feedback !

PR #200, opened 2023-01-24
[venuslimm] iP

venuslimm:
# Fake Duke
> I might be fake but my features are real!

## Welcome to Fake Duke 🧔 
Fake Duke is a Duke which can help you *remember all your tasks*. It's
- Easy to use
- Free of charge
- ***AN UPGRADED DUKE***

## Features 💯 
- [x] CLI version
- [x] Managing tasks
- [ ] GUI version

## How to use the latest Fake Duke? 🤔 
1. Download and setup [Java JRE](https://www.oracle.com/sg/java/technologies/downloads/) on your machine
2. Download the [JAR file](https://github.com/venuslimm/ip/releases/download/A-Jar/duke.jar) and double click on your machine. 
2.1. Or you can choose to run this command in your terminal: `java -jar <jar_file_name>.jar`
3. Do your tasks management!

### Actions on Fake Duke
- Add task
  - Todo: `todo <task description>`
  - Deadline: `deadline <task description> /by <YYYY-MM-DD HH:MM>`
  - Event: `event <task description> /from <YYYY-MM-DD HH:MM> /to <YYYY-MM-DD HH:MM>`
- Delete task: `delete <task index>`
- Mark task: `mark <task index>`
- Unmark task: `unmark <task index>`
- Find task: `find <keyword>`
- Exit: `bye`

Tasks are saved locally on your machine in `./data/tasks.txt`

## Creating your own Duke! 😄 
If you would like to make a Duke as well, you can get the skeleton code [here](https://github.com/nus-cs2103-AY2223S2/ip). This is how the original `main` method looks like:
```java
public class Duke {
                + "|  _ \\ _   _| | _____ \n"
                + "| | | | | | | |/ / _ \\\n"
                + "| |_| | |_| |   <  __/\n"
    }
}
```


zhacatomn (commented):
As far as I can tell, your code obeys the coding standard. LGTM!

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Exits from main program.
+     *
+     * @param tasks List of tasks.
+     * @param ui Ui object that handles all Ui actions.
+     * @param storage Storage object that handles all Storage actions.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        super.isExit = true;
```
seanfirefox: Not sure if this follows OOP guidelines, may consider using a setter method.

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,32 @@
...
+import duke.exception.DukeException;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Marks task when user input indicates mark.
+ */
+public class MarkCommand extends Command {
+    private final int TASK_INDEX;
```
seanfirefox: Consider using ```private static final int TASK_INDEX``` or ```private final int taskIndex``` as this might have caused warnings in checkstyle.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public String toString() {
+        Ui ui = new Ui();
+        return String.format("[E][%c] %s (from: %s to: %s) %s", this.getStatusIcon(), this.description,
+                ui.getStringDateTime(this.startDateTime), ui.getStringDateTime(this.endDateTime),
+                super.urgentMessage(this.startDateTime));
+    }
+
+    private void setEventDateTimes(String description) throws DukeException {
+        Parser parser = new Parser();
```
seanfirefox: Can consider setting all methods in parser to be static so there is no need to call the constructor.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,77 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    /**
+     * Marks task as done.
+     */
+    public void mark() {
+        this.isDone = true;
```
seanfirefox: Might want to refrain from unnecessary use of ```this```

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Handles interactions with the user.
+ */
+public class Ui {
+    private final Scanner SCANNER = new Scanner(System.in);
+
+    public String showLine() {
```
seanfirefox: Consider changing to a print statement rather than return to follow your naming convention (showWelcome(), showAdd() prints and does not return string)

seanfirefox (commented):
Overall code quality guidelines are well-followed. Good job!

src/main/java/duke/task/Event.java
venuslimm: Thank you for the suggestion! I have updated all the methods in Parser to be static.

src/main/java/duke/task/Task.java
venuslimm: Thanks for pointing it out. I have removed the unnecessary use of `this` in some parts of the code.

src/main/java/duke/command/MarkCommand.java
venuslimm: Resolved. Thanks for the suggestion!

PR #201, opened 2023-01-24
[anthonynmh] iP

anthonynmh:
# DukePro - Jarvis

   >“Your mind is for having ideas, not holding them.” – David Allen \([source](https://dansilvestre.com/productivity-quotes))
   
- text-based
* easy to learn
+ ~FAST~ _SUPER_ FAST to use 

All you need to do is,

2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :wink:

And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders

---

```java
public class Main {
    }
}
```

   ```
   Hello from
    ____        _        
   |  _ \ _   _| | _____ 
   | | | | | | | |/ / _ \
   | |_| | |_| |   <  __/
   |____/ \__,_|_|\_\___|
   ```


src/main/java/seedu/duke/Todo.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public Todo(int id, String task) {
+        super(id, task);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String printTask() {
```
venuslimm: I think the method name is slightly misleading. Would getTaskStatus be a better name?
```suggestion
    public String getTaskStatus() {
```

src/main/java/seedu/duke/ToDoList.java
```diff
@@ -0,0 +1,150 @@
...
+        todolist.remove(index - 1);
+        count--;
+    }
+
+    /**
+     * Marks the task at specified index as done.
+     *
+     * @param index index of task
+     */
+    public void markDone (int index) {
```
venuslimm: Should the extra space be removed?
```suggestion
    public void markDone(int index) {
```

src/main/java/seedu/duke/Task.java
```diff
@@ -0,0 +1,79 @@
+package seedu.duke;
+
+/**
+ * Handles the tasks relating to the ToDoList
+ */
+public class Task {
+
+    private int id;
+    private String task;
+    private boolean done;
```
venuslimm: Should this be renamed to `isDone`?
```suggestion
    private boolean isDone;
```

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Parses the user commands.
+     *
+     * @param command user command
+     * @param todolist the current ToDoList
+     * @param storage the storage associated with the current ToDoList
+     * @return isBye boolean indicated the ending 'bye' command
+     */
+    public boolean parse(String command, ToDoList todolist, Storage storage) {
```
venuslimm: I have noticed that you named your `ToDoList` objects as `todolist` in this line and other lines. Should the `todolist` variable rename to `toDoList`?
```suggestion
    public boolean parse(String command, ToDoList toDoList, Storage storage) {
```

venuslimm (commented):
LGTM! I have left some suggestions but generally I think your code is easy to read. I like how you separate the different logical units using lines.

src/main/java/seedu/duke/Todo.java
anthonynmh: Good point, thank you for the input!

src/main/java/seedu/duke/ToDoList.java
anthonynmh: Good catch!

src/main/java/seedu/duke/Task.java
anthonynmh: I agree, `isDone` makes more sense in this case. Thank you for the suggestion!

src/main/java/seedu/duke/Parser.java
anthonynmh: I believe that this depends on whether `todolist` is itself a Noun. However, I can see why it may cause confusion. Good catch!

PR #215, opened 2023-01-25
[Lavanya Mishra] iP

Lava-Iris:
Iris frees your mind of having to remember things you need to do. It is:
- text based
- easy to learn 
- ~~Fast~~ _SUPER_ FAST to use

All you need to do is:
1. download it from [here](https://github.com/Lava-Iris/ip/releases/tag/A-Jar)
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :wink:

And it is **FREE**!

Features:
- [x] Managing Tasks
- [x] Managing Deadlines
- [ ] Reminders (coming soon)

# 
If you are a Java programmer, you can use it to practice Java too. Here is the main method:

```
public class Main {
    }
}
```


src/main/java/iris/TaskStore.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * parses the stored file to return task list
+     * @return the task list stored in the file
+     * @throws IrisException if there is a problem in the stored file
+     */
+    public TaskList parse() throws IrisException {
+        TaskList tasks = new TaskList();
+        try {
+            try {
+                Scanner fileScanner = new Scanner(this.taskFile);
```
Merrickneo, line 87: Remember to close your scanner object!

src/main/java/iris/task/Event.java
```diff
@@ -0,0 +1,60 @@
...
+     * Constructor for Event
+     * @param name description of event
+     * @param from the start date/time of the task in the format: {dd-MM-yyyy HHmm}. Eg. 20-02-2002 2000
+     * @param to the end date/time of the task in the format: {dd-MM-yyyy HHmm}. Eg. 20-02-2002 2000
+     * @throws DateTimeException when there is a problem parsing given dates.
+     */
+    public Event(String name, String from, String to) throws DateTimeException {
+        super(name);
+        try {
+            this.from = LocalDateTime.parse(from.trim(), DateTimeFormatter.ofPattern("dd-MM-yyyy HHmm"));
```
Merrickneo: Just a suggestion but you could perhaps create another class like `DatetimeParser` to help with parsing of the datetime string!

src/main/java/iris/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+        case mark:
+            return new MarkTaskCommand(Integer.parseInt(input.split(" ")[1]) - 1);
+        case unmark:
+            return new UnmarkTaskCommand(Integer.parseInt(input.split(" ")[1]) - 1);
+        case delete:
+            return new DeleteTaskCommand(Integer.parseInt(input.split(" ")[1]) - 1);
+        case reset:
+            return new ResetCommand();
+        case filter:
+            try {
```
Merrickneo, line 107: You could make another subclass of Command to help handle filter commands like what you did above.

src/main/java/iris/Ui.java
```diff
@@ -0,0 +1,43 @@
...
+    public String readInput() {
+        return inputGetter.nextLine();
+    }
+
+    /**
+     * initial greeting to the user
+     */
+    public void greet() {
+        output(GREETING);
+    }
```
Merrickneo: Could have a `endProgram` function to help with the closing of your `Ui` when the user inputs `bye`.

Merrickneo (commented):
Hi Lavanya, excellent use of OOP to help with your program flow and logic.
Couldn't really pinpoint many major problems and your documentation is short and descriptive as well.

Keep it up!

src/main/java/iris/Parser.java
Lava-Iris: Oh, I do have a subclass of Command (FilterCommand). I initiate in on line 73. Here, I'm just trying to check the input, and separate it into two cases: when we want the deadlines and events on a date and when we need all deadlines and events in a certain period (both start and end date-time is provided). 

I have different constructors for both so I need to do it here. Also, OOP-wise, I think it's the work of Parser.

Thank you for the suggestion though. Do let me know if I misunderstood your suggestion.

src/main/java/iris/task/Event.java
Lava-Iris: Oh I was thinking of making a method for that (it's kinda messy and repetitive now), but I think a class might be better. Thank you!

src/main/java/iris/TaskStore.java
Lava-Iris: OH! I completely forgot about it. Thank you for the reminder! :)

PR #218, opened 2023-01-25
[engenhui1999] iP

engenhui1999:
# TaskGenie

TaskGenie frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use
- All you need to do is,

download it from [here](https://github.com/engenhui1999/ip/releases/tag/v1.0).

double-click it.
add your tasks.
let it manage your tasks for you 😉
And it is **FREE**! 😱 

Features:

- [X] Managing tasks
- [X] Managing deadlines
- [X] Managing events
- [ ] Reminders (coming soon)
---------------

```
public class Main {
        Launcher.main(args);
    }
}
```


src/main/java/Parser/Parser.java
```diff
@@ -6,7 +6,7 @@
 
 public class Parser {
     protected static enum CommandList {
```
kennethk-1201: I think your use of a command list is really neat!

src/main/java/Storage/TaskList.java
```diff
@@ -43,9 +43,26 @@ public boolean isEmpty() {
...
+    /**
+     * Returns all the matching tasks in string representation.
+     * @param description The description to be matched.
+     * @return The string representation of all matching tasks.
+     */
+    public String getMatchingTasks(String description) {
+        String str = "";
+        for (int i = 1; i <= this.db.size(); i++) {
+            if (this.db.get(i - 1).getDescription().contains(description)) {
+                str += (i + "." + this.db.get(i - 1) + "\n");
```
kennethk-1201: Could this line possibly be abstracted out as a function?

engenhui1999: Thank you for the suggestion! I will take note of this in my future commits!

src/main/java/Storage/TaskList.java
```diff
@@ -0,0 +1,94 @@
...
+        return this.db.size();
+    }
+
+    /**
+     * Checks if the database is empty.
+     * @return A boolean value based on whether the database is empty.
+     */
+    public boolean isEmpty() {
+        return this.db.isEmpty();
+    }
```
kynhan: I like the use of the "is" prefix for this boolean method.

src/main/java/Storage/TaskList.java
```diff
@@ -0,0 +1,94 @@
...
+    public int getSize() {
+        return this.db.size();
+    }
+
+    /**
+     * Checks if the database is empty.
+     * @return A boolean value based on whether the database is empty.
+     */
+    public boolean isEmpty() {
+        return this.db.isEmpty();
```
kynhan: Maybe instead of using this.db, you can just use db (like you did on line 29)? I think it would make the code neater and more concise.

kynhan (commented):
Overall, code is very neat and clean. 
Some minor things to consider only.

PR #222, opened 2023-01-25
[adam07018] iP

adam07018:
### **Duke**
`You may delay, but time will not. -Miles Davis`
### Always lost track of the deadline:scream:? Have a bad memory?:face_with_spiral_eyes:

Don't worry! Now you have Duke to help you. It can do following:
1. Add different kinds of task: Todo, Deadline, Event
4. Mark a task as done or not done yet
5. List out all the tasks you have written down
6. Delete any tasks from your list
7. Search the tasks you created by keywords

### Why is it so good?

- **Easy** to learn, there are examples provided for you:partying_face:
- Fast speed. **Run and Go**
- Record down your task lists into hardcopy
- **Free!!**
### How to use? Easy lah. Even my ah mah can do it!
- [ ] download it from [here](https://github.com/adam07018/ip).
- [ ] double-click it.
- [ ] add your tasks.
- [ ] And here we go!
### **Time doesn't wait!:star_struck: Download and use now to make your life much easier!**


src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,31 @@
+package duke;
+
+/**
+ * Exception class
+ * Print various messages to users due to different error situations
+ */
+public class DukeException extends Exception {
+    DukeException(ExceptionType e) {
+        switch (e) {
```
huanghao1998: I like you have used switch case to tackle different exception, but i would recommend to extend the different exceptions out into different exception class. That would be easier to print out exception message when implementing GUI.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,91 @@
+package duke;
+import java.util.ArrayList;
+
+/**
+ * Class that display messages (except error messages) to user
+ */
+public class Ui {
```
huanghao1998, line 8: The naming of the method is well done

src/main/java/duke/Times.java
```diff
@@ -0,0 +1,30 @@
+package duke;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+/**
+ * Object class for time
+ * Input of time will be formatted to MMM d yyyy hh:mm
+ */
+public class Times {
+    static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
```
huanghao1998, line 15: If ``FORMATTER`` is not used in other class and try to encapsulate it using ``private``

huanghao1998 (commented):
Overall, the code is neat, naming of method and class are well defined :)

src/main/java/duke/Times.java
adam07018: Noted with thanks

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,65 @@
...
+                cmd.execute(list);
+                if (cmd.isExit()) {
+                    sc.close();
+                    return;
+                }
+            } catch (DukeException e) {
+                continue;
+            } catch (NullPointerException e) {
+                continue;
+            }
```
ZHTang29: I like the use of the parser, and overall the OOP is done up appropriately! Clear java-docs as well so well done on the documentation portion.

src/main/java/duke/DukeException.java
ZHTang29: Likewise, I feel the exception handling could be further split into individual classes, but most exceptions are handled so well done on that.

ZHTang29 (commented):
Overall, I find the code easy to read, because you follow proper naming conventions and the javadocs in generally are done up well. Other than the minor portion on splitting the exception classes (if necessary to you), well done and keep up the good work!

PR #225, opened 2023-01-25
[sarthak181] iP

sarthak181:


src/main/java/Duke.java
```diff
@@ -6,5 +8,104 @@ public static void main(String[] args) {
...
+                    }
+                    list.add(c);
+                    System.out.println("Got it. I've added this task:\n  "
+                            + c.toString() + "\nNow you have " + list.size()
+                            + " tasks in the list.");
+                    break;
+                case 5:
+                    System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+
+            }
```
wz2k: cases in switch statements should not be indented. eg
```java
switch (variable) {
case 0:
    return 0;
case 1:
    return 1;
}

File: src/main/java/Duke.java [hunk 2]
```diff
@@ -6,5 +8,104 @@ public static void main(String[] args) {
...
+            System.out.println("" + i + ". " + t.toString());
+        }
+
+    }
+    static int getSwitch(String str) {
+        if (str.startsWith("mark ")) return 0;
+        if (str.startsWith("unmark ")) return 1;
+        if (str.startsWith("todo ")) return 2;
+        if (str.startsWith("deadline ")) return 3;
+        if (str.startsWith("event ")) return 4;
```
wz2k: Use K&R style brackets. eg
```java
if (condition) {
    return 0;
}
```
I suggest just using the string directly in the switch statement but this is entirely up to you.

src/main/java/Duke.java
```diff
@@ -1,3 +1,5 @@
+import java.util.*;
+
 public class Duke {
```
wz2k: Classes should be placed in a package. You can just place all your current class in a duke package minimally

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
+public class Event extends Task {
+    protected String from;
+    protected String to;
+
+    public Event(String description, String from, String to) throws DukeException{
```
wz2k: 'DukeException{'. Missed out on a space between n and {.

src/main/java/Duke.java
```diff
@@ -1,3 +1,5 @@
+import java.util.*;
```
wz2k: Wildcard imports should not be used. Imported classes should be listed explicitly

wz2k (commented):
Good code overall besides a few changes that should be made to follow our module's coding standard. Do remember to add descriptive header comments (eg. javadocs) to your public classes and methods. Keep up the good work! :thumbsup:

PR #228, opened 2023-01-25
[Jun-How] iP

Jun-How:
# JeoBot

_Struggling to keep your life organised?_
JeoBot is the right tool for you!

JeoBot keeps you worry-free by keeping track of the things you need to do.

### Plus Points ✔️
- Text-based
- User-friendly
- Convenient
- ~~FAST~~ BREAKNECK FAST to use
- What's more, it is **FREE**!

### Activation
1. Download the latest JAR release file from [here](https://github.com/Jun-How/ip/releases/tag/A-Release).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉

### Features
- [x] Managing different tasks (ToDo, Deadline, Event)
- [x] Finding tasks based on keyword or due date
- [x] Adding tags for customisation
- [x] GUI support provided
- [ ] Reminders (coming soon)

If you are a Java programmer, you may use it to practise Java too! Here's the main method:

```
public class Launcher {
    }
}
```
_What are you waiting for? Download it already!_


src/main/java/JeoBot.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Represents the bot which the user may run.
+ * @author Goh Jun How
+ * @version 0.1
+ */
+public class JeoBot {
+    protected static final String DATE_PARSE = "yyyy-MM-dd";
+    protected Ui ui;
+    protected Storage store;
+    protected TaskList taskList;
```
Diatbbin: Would it be better to set the access modifiers as private instead?

src/main/java/JeoBot.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Represents the list of commands recognised by the bot.
+     */
+    public enum Command {
+        BYE, LIST, MARK, UNMARK, DELETE, TODO, DEADLINE, EVENT, DUE, FIND
+    }
```
Diatbbin: Good use of enums!!

src/main/java/jeo/parser/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+                        ? "Please enter a task description."
+                        : "Please follow the format: event <description> "
+                        + "/from <yyyy-MM-dd HH:mm> /to <yyyy-MM-dd HH:mm>");
+            }
+            String[] arr = parseSubstringTasks(str, "event1");
+            String desc = arr[0];
+            str = arr[1];
+            arr = parseSubstringTasks(str, "event2");
+            String from = arr[0];
+            String to = arr[1];
```
Diatbbin: I think u can consider replacing the if-else structure with switch case for better code readability.

Diatbbin (commented):
:) 👍 Overall, really good use of OOP. Class structures seem to be well thought out. Coding style and variable namings looks good too! I think Parser class was abit hard to read, I think u can consider using functions to remove duplicated code.

src/main/java/jeo/parser/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+            hm.put("command", "find");
+            String key = str.substring(4).trim();
+            if (key.isEmpty()) {
+                throw new JeoException("Please enter a keyword.");
+            }
+            hm.put("key", key);
+        } else {
+            hm.put("command", "");
+        }
+        return hm;
```
ZHTang29: Similar to another comment on the same block of if-else code, with such a long block the readability is slightly compromised (and it is slightly harder to debug), so you may consider editing the code to cater to that (switch cases etc.).

ZHTang29 (commented):
Overall, coding quality is good and generally most portion of the code is readable and good use of OOP! Keep it up!

PR #229, opened 2023-01-25
[nus-jeff] iP

nus-jeff:
# D U K E #
> "*Duke is the best ~~partner~~ task tracker I have ever had!!*" - nobody ([hyperlink](https://www.youtube.com/watch?v=dQw4w9WgXcQ&ab&ab))
---
Start using Duke today, it is very **worth it** as it is:
+ Easy to use!
+ Super easy to use!!
+ Extremely easy to use!!!
---
Duke comes with the **ability** to:
1. Add or Delete tasks
2. Mark tasks as done or not done
3. Search for all tasks with a keyword
---
More features to be implemented:
- [x] Ability to exit Duke
- [ ] Ability to talk to Duke :kissing_heart: (coming soon!)
---
Here is a preview of our `Ui` in java code:
```Modelica
public static void welcomeMessage() {
            + "|  _ \\ _   _| | _____ \n"
            + "| | | | | | | |/ / _ \\\n"
            + "| |_| | |_| |   <  __/\n"
    System.out.println(DIVIDER_START_END
            + "What can the Duke help you with today?\n"
            + DIVIDER_START_END);
}
```


src/main/java/ToDoList.java
```diff
@@ -0,0 +1,133 @@
...
+        String divider = "@~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~@\n";
+        if (ind < 1 || ind > toDoCount) {
+            throw new IndexDukeException();
+        }
+        arr.get(ind).markDone();
+        System.out.println(divider + "The Duke has marked the following task: \n"
+                + " - " + arr.get(ind) + "\n" + divider);
+    }
+
+    public void save() throws IOException {
```
JamesLiuZX: Generally, for error handling, there are 2 types: adding a `throws` clause to a method, or using a try-catch block. The first one should only be used if you are throwing it upwards to be handled by the consumer method,  or caused by input parameters, or you don't want to handle them and just let the program stop. In all other cases, the try-catch block is preferred as you can meaningfully handle the exceptions and let users understand the error.

src/main/java/EventTask.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public String save() {
+        String status = this.isDone ? "DONE/+/" : "NOTDONE/+/";
+        return "EVENT/+/" + status + this.title + "/+/" + this.from + "/+/" + this.to + "\n";
+    }
+
+    @Override
+    public String toString() {
+        String status = this.isDone ? "[x] " : "[ ] ";
+        return "[E]" + status + this.title + " (from: " + this.from + " to: " + this.to + ")";
```
JamesLiuZX: Would be cleaner if you used String.format()

src/main/java/ToDoList.java
```diff
@@ -0,0 +1,133 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Scanner;
+import java.nio.file.Path;
+import java.util.regex.Pattern;
+
+
+public class ToDoList {
+    private ArrayList<Task> arr = new ArrayList<>(); //1-based index
```
JamesLiuZX: Generally, even for user inputs that start with 1-index. You should use 0-index for your internal data structures and convert the user inputs index. Otherwise you'll have to do a lot of necessary steps down the road with your methods such as deleting, checking for max length, and your constructor having to add a dummy task which takes up resources.

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+    private static ToDoList startUp() {
+        try {
+            String logo = " ____        _        \n"
+                    + "|  _ \\ _   _| | _____ \n"
+                    + "| | | | | | | |/ / _ \\\n"
+                    + "| |_| | |_| |   <  __/\n"
+                    + "|____/ \\__,_|_|\\_\\___|\n";
+            System.out.println("Hello from\n" + logo);
+            String divider = "____________________________________________________________\n";
+            System.out.println(divider + "What can the Duke help you with today?\n" + divider);
```
JamesLiuZX: If there's more than one instance, you can initialise the divider String as a private variable within your class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+                    default:
+                        throw new DukeException("The Duke does not understand your words!");
+                }
+            } catch (Exception e) {
+                System.out.println(e.getMessage());
+            }
+        }
+    }
+
+    private static String[] commandHandler(String input, String regex,int limit, int minSize) throws DukeException {
```
JamesLiuZX: Note proper spacing between arguments.

src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,36 @@
...
+    private LocalDate byDate;
+
+    public DeadlineTask(String title, String by) {
+        super(title);
+        this.by = formatIfDate(by, byDate);
+    }
+
+    private String formatIfDate(String input, LocalDate date) {
+        DateTimeFormatter inputFormat = DateTimeFormatter.ofPattern("d/M/uuuu");
+        DateTimeFormatter outputFormat = DateTimeFormatter.ofPattern("dd LLL uuuu");
```
JamesLiuZX: Would be clearer if you explained the use of 'L' and 'u' instead of M and Y

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+                        sub = input[1];
+                        ls.delete(Integer.parseInt(sub)); //numbersformatexception to be handled
+                        break;
+                    case "todo":
+                    case "event":
+                    case "deadline":
+                        Duke.taskCommandHandler(input, ls);
+                        break;
+                    default:
+                        throw new DukeException("The Duke does not understand your words!");
```
JamesLiuZX: Good job making use of exceptions in default!

JamesLiuZX: Could create a new specific exception for this.

JamesLiuZX: As per Java coding standards, the cases should be on the same indentation level as switch.

JamesLiuZX (commented):
Overall great work! The code logic is great, and I can see good use of OOP. Keep it up!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,50 @@
+package duke;
+
+import duke.functions.Parser;
+import duke.functions.Storage;
+import duke.functions.UI;
+
+import java.util.Scanner;
+
+public class Duke {
+    public static void main(String[] args) {
```
swxk19: I like how you abstracted out all the `switch-case` handling of the commands out of `main`. Makes it look short and sweet!

src/main/java/duke/ToDoList.java
```diff
@@ -0,0 +1,65 @@
...
+    public void add(Task task) {
+        ++this.toDoCount;
+        arr.add(task);
+    }
+
+    public Task delete(int ind) throws IndexDukeException {
+        if (ind < 1 || ind > toDoCount) {
+            throw new IndexDukeException();
+        }
+        Task rm = arr.get(ind);
```
swxk19: Variable name can be more descriptive, perhaps `taskToRemove`

src/main/java/duke/functions/Parser.java
```diff
@@ -0,0 +1,86 @@
...
+import duke.exceptions.DukeException;
+import duke.exceptions.InputDukeException;
+import duke.tasks.DeadlineTask;
+import duke.tasks.EventTask;
+import duke.tasks.Task;
+import duke.tasks.ToDoTask;
+
+public class Parser {
+    private  Parser() {
+    }
```
swxk19: Constructor may not need to be defined if you are just using the default constructor :)

swxk19 (commented):
Nice Job!

PR #234, opened 2023-01-25
[cjyothika] iP

cjyothika:
# Duke
> Why remember things? You're not going to do them anyway ~ Someone probably

Well, Duke is going to bring you closer to your goals, I guess *(still not completing them though)*. It's
- going to repeat what you tell it
- repeat what you say to it
- more repetition so I guess that helps, right?

All you need to do is,
1. download a release dummy (https://github.com/nus-cs2103-AY2223S2/ip)
2. run the following command in cli
`java -jar -insert file path-`
3. let it repeat your tasks back to you
4. Profit :heavy_dollar_sign:

And it involves **minimal** effort?

## Features:
- [X] repeating tasks back to you
- [ ] was i supposed to think of more features? oops

If you're a Programmer, I guess here's some code to read:
```
System.out.println("Hello world");
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+        } catch (DukeException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Run.
+     */
+    public void run() {
```
yaojiax, line 36: Could possibly use protected here instead of public

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,84 @@
...
+        this.description = description;
+        this.from = from;
+        this.to = to;
+        this.type = "E";
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        switch (this.type) {
+            case "E":
```
yaojiax: switch and case should have the same indentation, i.e. they should start from the same line. No worries, this is a common mistake.

src/test/java/duke/command/ParserTest.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Tests if Parser parses deadline commands correctly.
+     *
+     * @throws DukeException the duke exception
+     */
+    @Test
+    public void parseDeadlineTest() throws DukeException {
+        String[] actual = Parser.processDeadline("deadline hw /by 2023-01-01");
+        String[] expected = new String[]{" hw ", "2023-01-01"};
+        assertEquals(expected.length, actual.length);
```
yaojiax, line 35: It's clear as to what you are testing. Nice!

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import duke.Storage;
+import duke.DukeException;
+
+/**
+ * The type Mark command.
+ */
+public class MarkCommand extends Command{
+
+    private int index;
+    private boolean isMarked;
```
yaojiax, line 14: Good naming of Boolean variables!

yaojiax (commented):
Nice and readable code with just a few coding style violation. Keep it up!

src/main/java/Duke.java
cjyothika: True. Thanks for the suggestion. However, I was considering the case where a user might want to call run() again on an already created Duke object.

src/main/java/duke/command/AddCommand.java
cjyothika: Thanks for this suggestion. I missed that. :)

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,128 @@
+package duke.command;
+
+import duke.DukeException;
+
+/**
+ * The type Parser.
+ */
+public class Parser {
+
+    public enum Keyword {
```
Cheamybunny, line 13: Good use of ENUMS

Cheamybunny (commented):
Looks good so far, look forward to looking at your GUI  implementation

PR #235, opened 2023-01-25
[bokun] iP

bokung:
UwU
- *OwO*
- **OwO OwO**
- OwO OwO OwO

(>w<)
1. ^w^ :robot:
2. ^w^ :robot:
3. ^w^ :robot:

YwY
- [x] ~~UwU~~ VwV
- [ ] VwV
- [ ] VwV

Here is some code to perform a `binarySearch`, but there's a bug hidden somewhere! Can you spot what's wrong?
``` java
public class BinarySearch {
    public static int binarySearch(int[] elements, int target) {
        int begin = 0;
        int end = elements.length - 1;
        while (begin < end) {
            int mid = begin + (end-begin)/2;
            if (target <= elements[mid]) {
                end = mid;
            } else {
                begin = mid;
            }
        }
        return elements[begin] == target ? begin : -1;
    }
}
```


src/main/java/task/Task.java
```diff
@@ -8,27 +8,24 @@ public Task(String taskName) {
...
-
-    public String getTaskNameWithStatusIcon() {
-        return String.format("%s %s", this.getStatusIcon(), this.getTaskName());
+    @Override
+    public String toString() {
+        return String.format("%s %s", this.getStatusIcon(), this.taskName);
     }
 
     public String markDone() {
         this.markedDone = true;
```
russelltankaimin: I think a better naming for `markedDone` boolean variable would be `isDone`. According to the textbook, some of the recommended prefixed are `is`, `has` etc. For your consideration 😊

src/main/java/Duke.java
```diff
@@ -28,9 +28,7 @@ private static void bye() {
     private static void displayTasks(Task[] tasks, int numTasks) {
         StringBuilder sb = new StringBuilder();
```
russelltankaimin: I might be a bit nitpicky, but perhaps a better naming for this String Building could just simply be `stringBuilder` rather than `sb`. For your consideration 😊

src/main/java/task/Task.java
```diff
@@ -8,27 +8,24 @@ public Task(String taskName) {
...
-    }
-
-    public String getTaskNameWithStatusIcon() {
-        return String.format("%s %s", this.getStatusIcon(), this.getTaskName());
+    @Override
+    public String toString() {
+        return String.format("%s %s", this.getStatusIcon(), this.taskName);
     }
 
     public String markDone() {
```
russelltankaimin: Yes, one good habit in your coding style is that you name methods starting with verbs! Great job! 👍

russelltankaimin (commented):
Great job overall with your overall coding style. I do not see any major issues. You followed almost everything from the coding style guide.

src/main/java/Duke.java
```diff
@@ -18,7 +18,7 @@ public class Duke {
 
     private final static int CAPACITY = 100;
```
russelltankaimin: I think the order should have been `private static final` instead.

src/main/java/Duke.java
```diff
@@ -94,14 +96,24 @@ public static void main(String[] args) {
                         break;
                     case MARK:
```
russelltankaimin: Do make sure, your `switch` and `case` keywords lie under the same indentation line 👌

src/main/java/task/Task.java
bokung: Good catch, will update in next version!

src/main/java/Duke.java
bokung: Will consider! Though for this one the scope is rather limited and this stringbuilder is not really being used outside of this function.

src/main/java/Duke.java
bokung: Will take note.

src/main/java/Duke.java
bokung: Good catch. Will update in future version.

src/main/java/uwuke/input/Advisor.java
```diff
@@ -0,0 +1,26 @@
...
+ * TODO: Add more functionality to the class in the future, advise is not that helpful currently.
+ */
+public class Advisor {
+    /**
+     * Advise users on potential fixes
+     * 
+     * @param input faulty input string given by user
+     * @return advise based on faulty input
+     */
+    public static String advise(String input) {
```
panavdua: Perhaps here you could make the name of the function more indicative of what it is doing, for example - `adviseUsers`

src/main/java/uwuke/input/Parser.java
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Processes string into useful information when command is identified as a todo command
+     * <p> Regular Expression: ^todo\\s.+$
+     * 
+     * @param input todo command that matches the above regular expression
+     * @return string that can be used directly to create a new deadline object
+     * @throws DukeException if input cannot be parsed
+     */
+    public static String parseTodo(String input) throws DukeException {
```
panavdua: Very well-written function!! But wanted to ask if you would prefer to keep `Todo` or change it to `ToDo`?

panavdua (commented):
You really have adhered closely to the coding standard, could not find any major error!!

src/main/java/uwuke/input/Parser.java
bokung: Yes it should be `ToDo`. Good catch!

src/main/java/uwuke/input/Advisor.java
bokung: Will do that in the next version! It does indeed appear that this entire class is a bit ambiguous and can be named to reflect its purpose better.

PR #236, opened 2023-01-25
[anshumaantgi] iP

anshumaantgi:
# Ava
>  Why forget , when Ava can keep track of it ? - The Creator of Ava

Ava frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn 
- ~~FAST~~ SUPER FAST to use 

All you need to do is ,
1. Download it from [here](https://github.com/anshumaantgi/ip)
2. double-click it.
3. add you tasks.
4. let it manage your task for 😉 

And it's **FREE**

Features:
- [x] Managing Task
- [ ] Managing deadlines
- [ ] Reminders (coming soon)

if you're a Java programmer, you can. use it to practice Java too. Here's the `main` method:
```java
public class Main {
    }
}
```


src/main/java/Ava/Commands/Exit.java
```diff
@@ -0,0 +1,27 @@
...
+     */
+    @Override
+    public boolean run(TaskList t, Storage s){
+        return false;
+    }
+
+
+    @Override
+    public String output(String formatSpace){
+        return "Will Never reach here";
```
Tempura-Person: Should probably be a comment instead of a String return?

src/main/java/Ava/Commands/Exit.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     *
+     * @param t a TaskList object
+     * @param s a Storage object
+     * @return False to indicate program has stopped runnign
+     */
+    @Override
+    public boolean run(TaskList t, Storage s){
+        return false;
```
Tempura-Person: If all the run method does is return false, and no special 'goodbye' message is printed here, is this class necessary?

src/main/java/Ava/Commands/FindTask.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    @Override
+    public boolean run(TaskList t, Storage s) throws AvaException {
+        this.keyword = this.parsedInput[0];
+        this.tasks = t;
+        return true;
+    }
+
+    /**
+     *
```
Tempura-Person: Seems to be an extra newline here in the javadocs.

src/main/java/Ava/Exceptions/AvaException.java
```diff
@@ -0,0 +1,6 @@
+package Ava.Exceptions;
```
Tempura-Person: I like how you made several Exception classes to handle different exception cases.
However, they do seem a little all over the place in terms of the naming convention.
(e.g. CommandNotFoundException should probably be named CommandNotFound, to fit with the other ones.)

src/main/java/Ava/Exceptions/CommandNotFoundException.java
```diff
@@ -0,0 +1,14 @@
+package Ava.Exceptions;
+
+public class CommandNotFoundException extends AvaException {
+    private final String ERROR_PROMPT = "Command Not Complete or Possible !!";
```
Tempura-Person: The other Exception classes seem to give a clear indicator of what the correct way is for the user to input text.
Perhaps this Exception could tell the user the correct command formats?

Tempura-Person (commented):
Quite a solid IP, there was a good amount of thought put into structuring the individual Classes.
However, I am not sure why there is a duplicate directory for Ava.
(there is both an Ava file and an ava file, containing Class duplicates.)

anshumaantgi:
> Quite a solid IP, there was a good amount of thought put into structuring the individual Classes. However, I am not sure why there is a duplicate directory for Ava. (there is both an Ava file and an ava file, containing Class duplicates.)

> Thanks for the feedback , I have removed the two folders and just left the one .

PR #239, opened 2023-01-25
[dsja612] iP

dsja612:
# Duke
> "Quotations on the Internet are often misattributed." - Duke

# Yet another school project??
Duke is a task manager written in Java, in an object-oriented manner. This is all so you can live your life in a task-oriented manner 🤓

# "Pfft, why would I use Duke instead of Google Calendar??" - 🤓
- Text-based, specifically for tech 🤓s.
- Easy to learn
- As fast as you can type

And it is **FREE!**

# Features
- [x] CLI based (for now)
- [x] Saves your tasks even after you exit!
- [x] Everything is abstracted away just like OOP!

Download it [here!](https://github.com/dsja612/ip/releases/tag/A-Jar)

## If you want to learn Java & OOP:
Feel free to check out how Duke is organized. Duke starts running from this code block in `Duke.java`:
```ruby
    public static void main(String[] args) throws IOException {
        Duke.run();
    }
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+    private static final TaskList tasks = new TaskList();
+    private static final Scanner sc = new Scanner(System.in);
+    private static final UI ui = new UI();
+    private static final Storage storage = new Storage(tasks);
```
immanuelhume: Since these variables are static, perhaps it is best to make the constructor for the `Duke` class private, and make it a final class? This prevents multiple Duke instances from being created.

src/main/java/commands/EventCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    private LocalDateTime endDate;
+
+    public EventCommand(String name, String startDate, String endDate) throws DukeException {
+        this.name = name;
+        try {
+            this.startDate = LocalDateTime.parse(startDate, DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm"));
+            this.endDate = LocalDateTime.parse(endDate, DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm"));
+            if (this.startDate.isAfter(this.endDate)) {
+                throw new DukeException("The starting date of this event is after its ending date!");
+            }
```
immanuelhume: Maybe we can move this `if` block out of the `try-catch` block, to reduce the amount of nesting.

src/main/java/commands/FindCommand.java
```diff
@@ -0,0 +1,38 @@
...
+    private String searchTerm;
+
+    public FindCommand(String searchTerm) {
+        this.searchTerm = searchTerm;
+    }
+
+    @Override
+    public void execute(UI ui, TaskList tasks, Storage storage) throws DukeException {
+        if (tasks.size() == 0) {
+            this.commandStatus = "You have no tasks in your list! \n";
```
immanuelhume: Could consider adding an early `return` here, then we can drop the else block.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.FileNotFoundException;
+
+/**
+ * Handles loading and saving of tasks to local storage.
+ */
+public class Storage {
+    private final Path HOME_DIRECTORY = Path.of(System.getProperty("user.dir") + "/data");
```
immanuelhume: I guess this variable name might be misunderstood to mean `$HOME` on unix. Would `DUKE_DATA_DIR` be better?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,109 @@
...
+
+        switch (taskType) {
+            case "T":
+                tasks.addTask(new ToDo(taskName));
+                break;
+            case "D":
+                String by = inputList[2];
+                LocalDateTime deadline = LocalDateTime.parse(by, DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm"));
+                tasks.addTask(new Deadline(taskName, deadline));
+                break;
```
immanuelhume: Regarding variable declarations in `case` blocks, it might be better to wrap them inside their own scope. This prevents potential name clashes with other branches of the `switch` statement.

```suggestion
            case "D": {
                String by = inputList[2];
                LocalDateTime deadline = LocalDateTime.parse(by, DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm"));
                tasks.addTask(new Deadline(taskName, deadline));
                break;
            }
```

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public String toString() {
+        StringBuilder result = new StringBuilder();
+        result.append("List of Tasks: \n");
+        for (int i = 0; i < taskList.size(); i++) {
+            result.append((i + 1));
+            result.append(". ");
+            result.append(taskList.get(i).toString());
+            result.append("\n");
```
immanuelhume: Could consider chaining methods here so it's less verbose.

```suggestion
            result.append((i + 1))
                    .append(". ")
                    .append(taskList.get(i).toString())
                    .append("\n");
```

immanuelhume (commented):
Nice work bro :+1:

src/main/java/storage/Storage.java
dsja612: Great observation, thanks! I wasn't sure of whether to add braces for every block as that would further increase the indentation of the code

resolved

src/main/java/views/UI.java
```diff
@@ -0,0 +1,57 @@
...
+        System.out.println("I hope you've managed to be productive today. Bye!");
+    }
+
+    private String indentString(String s) {
+        return "\n\t" + s;
+    }
+
+    /**
+     * Displays a list of possible commands
+     * to the user
```
sumhungyee: Might it be better to end with a fullstop?

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,45 @@
...
+    private LocalDateTime startDate;
+
+    /** The ending date of this task. **/
+    private LocalDateTime endDate;
+
+    /**
+     * Constructor for an Event task.
+     * @param name Name of the task
+     * @param startDate Starting date of the task
+     * @param endDate Ending date of the task
```
sumhungyee: Might it be better to add a fullstop behind these lines?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+        String taskName = inputList[1];
+
+        switch (taskType) {
+            case "T": {
+                tasks.addTask(new ToDo(taskName));
+                break;
+            }
+            case "D": {
+                String by = inputList[2];
+                LocalDateTime deadline = LocalDateTime.parse(by, DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm"));
```
sumhungyee: Is this line a bit too long?

sumhungyee:
Great job dude 👍

sumhungyee (commented):
Overall, I find your code easy to read! For the most part, it looks like coding standards were upheld! Great Job!

src/main/java/exceptions/DukeException.java
```diff
@@ -0,0 +1,15 @@
+package exceptions;
+
+/**
+ * General exception class for displaying error messages to users.
+ */
+public class DukeException extends Exception {
+
+    /**
+     * Constructor for DukeException that stores an error message.
+     * @param message Main body of error message
```
sumhungyee: Perhaps this needs capitalising as well!

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+    private File dataFile;
+    private final TaskList tasks;
+
+    public Storage(TaskList tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * Loads tasks from a local file into a TaskList, creates one if one does not exist.
+     * @throws IOException If fail to load/create a storage file
```
sumhungyee: Perhaps this needs capitalising as well!

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Parse input from file into tasks.
+     * @param input Line from local file
+     * @param tasks TaskList to store tasks
+     * @throws DukeException If fail to create task from invalid input
+     */
+    public static void parseFile(String input, TaskList tasks) throws DukeException {
+        String[] inputList = input.split(",");
+        String taskType = inputList[0];
+        String taskName = inputList[1];
```
sumhungyee: Your code here looks extremely readable! 👓

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+            case "T": {
+                tasks.addTask(new ToDo(taskName));
+                break;
+            }
+            case "D": {
+                String by = inputList[2];
+                LocalDateTime deadline = LocalDateTime.parse(by, DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm"));
+                tasks.addTask(new Deadline(taskName, deadline));
+                break;
+            }
```
sumhungyee: The case statements are at a wrong indentation level! Perhaps you may need to shift them down one level!

sumhungyee (commented):
Well done!  Your code looks fantastic!

src/main/java/storage/Storage.java
dsja612: Good catch!

PR #243, opened 2023-01-25
[vaidyanaath] iP

vaidyanaath:
# Duke
> "For every minute spent organizing, an hour is earned." -Anonymous

Say goodbye to cluttered thoughts and to-do lists. Duke is the ultimate solution for leading a more organized life.
Add tasks, manage them easily and keep track of upcoming deadlines or events.

It's
- text-based
- user-friendly
- ~~FAST~~ *BLAZINGLY* FAST ⚡ 

Follow these steps to get started:
1. download it from [here](https://github.com/vaidyanaath/ip/releases)
2. double-click it
3. add your tasks
4. sit back and manage your tasks with ease 😎

And it is absolutely **FREE!** 💸 

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [ ] Reminders (coming soon!)

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method:
```java
    start();
}
```


src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+import duke.task.Task;
+import duke.task.ToDo;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.exception.DukeStorageException;
+
+
+public class Storage {
+    private static final Path dataPath = Paths.get("data");
+    private static final Path taskListPath = Paths.get("data", "duke.txt");
```
cydtseng: Would be good to change constant names to adhere to UPPER_SNAKE_CASE.

cydtseng (commented):
LGTM! :+1:
Suggested minor change.

src/main/java/duke/storage/Storage.java
vaidyanaath: Thank you for taking the time to review my code and for bringing up this naming convention suggestion. I appreciate the suggestion and have made the necessary changes.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,187 @@
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+
+public class TaskList {
+    private final ArrayList<Task> TASK_LIST;
```
runoutofit: i like that you wrote the constant as UPPER_SNAKE_CASE

src/main/java/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+        String command = userCommand.split(" ", 2)[0];
+
+        // Single word commands
+        if (userCommand.equals("list")) {
+            tasks.printTaskList();
+            return;
+        }
+
+        // Multi-word commands
+        switch (command) {
```
runoutofit: i like that you implement switch case instead of multiple if-else statements which i see more often. i like how concise it makes your code look

src/main/java/Deadline.java
```diff
@@ -1,14 +1,21 @@
 public class Deadline extends Task {
 
-    private String by;
+    private final String dueDate;
```
runoutofit: since dueDate is a constant, would upper snake case be more suitable to adhere to coding standard?

src/main/java/Ui.java
```diff
@@ -0,0 +1,23 @@
+public class Ui {
+    private final String LOGO = " ____        _        \n"
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___|\n";
+
+    public void greet() {
+        String greeting = "____________________________________________________________\n"
```
runoutofit: would it be better to abstract out this line as a variable since you are using it often?

runoutofit (commented):
overall code is very good

PR #245, opened 2023-01-25
[kimberlybp] iP

kimberlybp:
# Sirius
> "Time is the scarcest resource and unless it is managed nothing else can be managed." - Peter Drucker ([source](https://www.brainyquote.com/quotes/peter_drucker_122188))

Sirius is a CLI app that helps you manage your time. It's,
- text-based
- easy to use
- ~~FAST~~ SUPER FAST to use

## In order to use it..
1. Download the JAR file from [here](https://github.com/kimberlybp/ip/releases/download/A-Jar/ip.jar)
2. Double-click it
3. Add your tasks
4. **Let Sirius handle the rest!** 😎

## Features
- [x] List all your tasks
- [x] Manage Todos, Deadlines and Events
- [ ] Reminders (Coming soon!)

### Devs feel free and play around with the code
Here's a preview of the `main` method: 
```java
    new Duke(PATH).run();
}
```


src/main/java/commands/CommandType.java
```diff
@@ -0,0 +1,23 @@
...
+    UNKNOWN("unknown"),
+    BYE("bye");
+
+    private final String value;
+
+    CommandType(String value) {
+        this.value = value;
+    }
+
+    public boolean equals(String input) {
```
Skeltons: Perhaps the methods should be called `isEqual`?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,72 @@
+package duke;
+
+import commands.*;
+import exceptions.DukeException;
+import tasks.*;
```
Skeltons: Shouldn't wildcard imports be avoided?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
+package duke;
+
+import tasks.*;
+
+import java.io.*;
```
Skeltons: Shouldn't wildcard imports be avoided?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+            switch (type) {
+                case "T":
+                    initList.add(new Todo(taskDesc, isCompleted));
+                    break;
+                case "D":
+                    initList.add(new Deadline(taskDesc, isCompleted, strArr[3]));
+                    break;
+                case "E":
+                    initList.add(new Event(taskDesc, isCompleted, strArr[3], strArr[4]));
+                    break;
```
Skeltons: Shouldn't there be no indentation between switch and case?

Skeltons (commented):
I've noticed that your Ui class is not being used properly as some classes are still printing to console without calling the Ui class. Overall, your code looks good to me!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,49 @@
...
+
+import commands.Command;
+import exceptions.DukeException;
+
+import java.io.File;
+import java.time.format.DateTimeParseException;
+import java.util.Scanner;
+
+public class Duke {
+    private static final String PATH = "data/Duke.txt";
```
jinbesan: Nice use of capitalization for constants

src/main/java/commands/CommandType.java
jinbesan: I think `isEqual` is a good idea!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return CommandType.UNKNOWN;
+    }
+
+    public static boolean isDescriptionEmpty(String[] arr) {
+        return arr.length == 1 || arr[1].trim().isEmpty();
+    }
+
+    public static Command parse(String input) throws DukeException {
+        String[] inputArr = input.split(" ", 2);
```
jinbesan: Perhaps we can better understand this String array variable by naming it `inputs` or `inputTokens`?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+                }
+                taskInput = inputArr[1];
+                Todo todo = new Todo(taskInput);
+                return new AddCommand(CommandType.TODO, todo);
+            case DEADLINE:
+                if (isDescriptionEmpty(inputArr)) {
+                    throw new DukeException("Hey now.. The description of a deadline cannot be empty. >:(");
+                }
+                taskInput = inputArr[1];
+                String[] dArr = taskInput.split("/by", 2);
```
jinbesan: The naming of this array might not be immediately evident on what it is supposed to be, because of the short name. Perhaps we can use a more precise name?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class Storage {
+    private final String filePath;
+
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public List<Task> loadFile() throws Exception {
+        List<Task> initList = new ArrayList<>();
```
jinbesan: Perhaps we can rename this Collection variable to be `tasks` as it is a collection of Tasks?

PR #249, opened 2023-01-25
[markusyeo] ip

markusyeo:
**Not Duke, but Joe**
> “You shouldn't work so hard, you should sleep early.” – Prof Wai Kay(CS1010S)

Remember this is not Duke. but Joe! Joe is your personal butler in your pocket who can remember what you have to do.

- text-based
- somewhat easy to learn
- just regularly FAST to use

All you need to do is,

download it from somewhere on my git page
2. add your tasks.
3. let it manage your tasks for you 😉
And it is **FREE** (probably the only good thing about it)!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

```
        ...
        String logo = "     |             \n"
                           + "     |  _ \\    _ \\ \n"
                          + " \\   | (   |   __/ \n"
                           + "\\___/ \\___/  \\___|\n";
```


src/main/java/duke/Duke.java
```diff
@@ -1,8 +1,17 @@
...
+ * This is Duke who initialises Joe the Java Bot!
+ */
 public class Duke {
 
+    /**
+     * Runs Joe, my java bot!
+     *
+     * @param args Input arguments, will be ignored.
+     */
     public static void main(String[] args) {
```
Yaladah, line 29: Maybe you could consider abstracting out the behaviour of the bot into a method run()?

Yaladah (commented):
I like that I could find basically no coding violations in the code. It might be nice to abstract out a Ui class that handles interaction with the user, such as printing of messages and taking in inputs from the user. (It seems to be a requirement for the ip too if I'm not mistaken.)

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.time.LocalDateTime;
+
+/**
+ * Represents an Event entered by the user.
+ */
+public class Event extends Task {
+
+    private final LocalDateTime fromTime;
+    private final LocalDateTime toTime;
```
gremmyz, line 11: Good job with the descriptive naming!

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,41 @@
...
+        try {
+            if (s.equals("ToDoException")) {
+                throw new ToDoException();
+            } else {
+                throw new UnknownCommandException();
+            }
+        } catch (Exception e) {
+            throw e;
+        }
+    }
```
gremmyz: Really cool way of handling exceptions 👍🏼

src/main/java/duke/Duke.java
gremmyz: Think I agree with this, I believe some lines here can be abstracted out into the Joe class.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,136 @@
...
+            return mark(input);
+        } else if (input.startsWith("unmark")) {
+            return unmark(input);
+        } else if (input.startsWith("delete")) {
+            return delete(input);
+        } else if (input.startsWith("find")) {
+            return find(input);
+        } else {
+            return addTask(input)[1];
+        }
```
gremmyz: Out of curiousity, any reason why you chose to use _if-else_ here instead of _switch_ like below?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,136 @@
...
+            case "event": {
+                String[] s = split[1].split(" /from ", 2);
+                String taskDescription = s[0];
+                s = s[1].split(" /to ");
+                LocalDateTime fromTime = LocalDateTime.parse(s[0], formatter);
+                LocalDateTime toTime = LocalDateTime.parse(s[1], formatter);
+                Task task = new Event(taskDescription, fromTime, toTime);
+                taskList.add(task);
+                return new String[]{String.valueOf(taskList.size()), String.format("added: %s", task)};
+            }
```
gremmyz: Any reason why you use one method to handle addTask inputs from both the CLI and the saved file?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,136 @@
...
+        task.mark();
+        return String.format("\tNice! I've marked this task as done:\n\t  %s", task);
+    }
+
+    private String unmark(String input) {
+        int i = Integer.parseInt(input.split(" ", 2)[1]) - 1;
+        Task task = taskList.remove(i);
+        task.unmark();
+        return String.format("\tNice! I've marked this task as not done yet:\n\t  %s", task);
+    }
```
gremmyz: Not sure if it's a missing access modifier for `mark` so I thought I'd let you know in case.

gremmyz (commented):
LGTM! Overall, I think your code is really clean, good job! Just missing a few javadocs here and there, may want to look into it if you have time! 😄

PR #253, opened 2023-01-25
[Vigneshwar Hariharan] iP

vigonometry:
# Leo - A CLI-based Productivity System
> Hola, me llamo Leo - _"Hello I am Leo"_

Leo frees your mind to focus on the things you **care about**. It is,

- text-based
- interactive
- ~Fun~ _Super Fun_ and,
- INCREDIBLY FAST to use.

## Features:
- **Three task types: Todo, Event and Deadline**
You can now create Todo, Event and Deadline items by prefacing the task content with todo, event or deadline.
Event and Deadline types require further description: `event ... /from DD/MM/YYYY HHMM /to DD/MM/YYYY HHMM`, `deadline ... /by DD/MM/YYYY HHMM`

- **Added task management functions: list, delete, mark as done and unmark completed tasks**
`list` displays all tasks currently saved in your Task List.
`delete <taskId>` deletes task with taskId in your Task List.
`mark <task Id>` and `unmark <taskId>` can be used to mark a task as completed or incomplete.

- **Added task search functions: find**
`find <keyword>` finds all tasks in your Task List that contains the keyword.

## Setup
1. You can download it from [here](https://github.com/vigonometry/ip/releases/tag/A-Jar)
2. Run it on your terminal using `java -jar leo.jar` (Requires `Java 11` or later).
3. Enter your tasks.

It's as **easy** as 1, 2, 3. 

---
If you're a novice programmer, you can learn Object-Oriented Programming using Java too! Here's the `main` method:

```
public static void main(String[] args) throws LeoTaskException {
        new Leo().start();
}
```

- [ ] If you enjoyed my bot, check this box 😉.


src/main/java/leo/task/Task.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public void setDone() {
+        isDone = true;
+        System.out.println(
+                "Well done on completing the task! Let me mark that as done! Campeon del mundo!");
+        System.out.println(this);
+
+    }
+
+    public void resetDone() {
```
Nicklelodeon: The method name resetDone() might be confusing, as it could be interpreted as reset to done. Perhaps setAsUndone() could be used?

src/main/java/leo/task/LeoTaskException.java
```diff
@@ -0,0 +1,42 @@
...
+    MissingTimelineException() {
+        super("Not this again, indicate a from and to. I'm as happy as I can be—but I have been happier.\n");
+    }
+}
+
+class EmptyDeletionException extends LeoTaskException {
+    EmptyDeletionException() {
+        super("Bruh, why are you trying to delete from an empty task list...\n");
+    }
+}
```
Nicklelodeon: I like the fun chat interactions you have created between the bot and the user, especially the cold rainy night in stoke reference :)) Creating specific exception classes that inherit from LeoTaskException is also a great approach!

src/main/java/leo/task/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+                case "unmark":
+                    tasks.get(Integer.parseInt(rqDetails[1]) - 1).resetDone();
+                    break;
+                case "delete":
+                    deleteTask(Integer.parseInt(rqDetails[1]));
+                    break;
+                default:
+                    addTask(request);
+                    break;
+            }
```
Nicklelodeon: Should switch and case have the same level of indentation?

src/main/java/leo/task/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+            tasks.add(task);
+            System.out.printf("Alright, I've added: %s to your task list.\n", tasks.get(tasks.size() - 1));
+            System.out.printf("You have %d tasks in your list, vamos, get moving!\n", tasks.size());
+        } catch (LeoTaskException e) {
+            e.printStackTrace();
+            System.out.println();
+        }
+    }
+
+    private void deleteTask(int taskId) throws LeoTaskException {
```
Nicklelodeon, line 34: I like how you lower cased the abbreviation in the variable name taskId, as you avoided the common mistake of naming it as TaskID instead!

Nicklelodeon (commented):
Overall, I felt that your naming conventions for methods and variables were really well done. Most of the coding conventions were adhered to too! I also liked your interesting chat UI interactions between the user and chat bot. 
I believe you could include more OOP to improve readability and abstraction in your code.
Some cosmetic changes were suggested!

src/main/java/leo/task/Task.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final long serialVersionUID = 1L;
+    
+    enum Type {
+        TODO, DEADLINE, EVENT
+    };
+
+    private String taskDesc;
+    private boolean isDone = false;
+    private Type type;
+    public final static List<String> descCommands = Arrays
```
francisyzy: Hi, I think this variable name can be more descriptive, when I first saw it I thought it was short for describeCommands but it seems like it is short for descriptionCommands.

Was thinking you can put your commands as an enum, with a tag to check if it has any descriptions.
Something like:
```java
public enum Command {
    TODO("todo", true),
    LIST("list", false);

    private final String command;
    private final boolean desc;

    Command(String command, boolean desc) {
        this.command = command;
        this.desc = desc;
    }

    public String getCommand() {
        return command;
    }

    public int getDesc() {
        return desc;
```

src/main/java/leo/task/Task.java
```diff
@@ -0,0 +1,201 @@
...
+
+    public void setNotDone() {
+        isDone = false;
+        Ui.printDivider();
+        System.out.println("Ok, I've marked that as not done! Please get to it :(");
+        System.out.println(this);
+        Ui.printDivider();
+    }
+
+    public boolean getDone() {
```
francisyzy: Good variable naming for the `isDone` and their respective methods!

src/main/java/leo/task/Task.java
```diff
@@ -0,0 +1,201 @@
...
+import java.io.Serializable;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalDateTime;
+import java.util.Arrays;
+import java.util.List;
+
+import leo.ui.Ui;
+
+public class Task implements Serializable {
+    private static final long serialVersionUID = 1L;
```
francisyzy, line 17: Hey, I think this needs to be in all caps?

> A constant can be named using ALL_CAPS.

src/main/java/leo/task/Task.java
```diff
@@ -0,0 +1,201 @@
...
+        TODO, DEADLINE, EVENT
+    };
+
+    private String taskDesc;
+    private boolean isDone = false;
+    private Type type;
+    public final static List<String> descCommands = Arrays
+            .asList(new String[] { "todo", "deadline", "event", "mark", "unmark", "delete" });
+
+    public final static List<String> commands = Arrays
```
francisyzy: Similarly, for commands I think it should be all caps?

`Variable 'commands' should be in ALL_CAPS (if it is a constant) or be private (otherwise). [ConstantName]`

francisyzy (commented):
I like it how you named your variables! most of it looks good other than a few minor things.

src/main/java/leo/task/Task.java
vigonometry: Hi Francis, thanks for your comments. Pertaining to this field, the JavaDocs for implementing a `Serializable` object prescribes the field be named as such.

> The serialization runtime associates with each serializable class a version number, called a serialVersionUID, which is used during deserialization to verify that the sender and receiver of a serialized object have loaded classes for that object that are compatible with respect to serialization...:
 `ANY-ACCESS-MODIFIER static final long serialVersionUID = 42L;`

src/main/java/leo/task/Task.java
vigonometry: Thanks for raising this up. Accidentally changed the access modifiers while implementing more classes.

src/main/java/leo/task/Task.java
vigonometry: This is fantastic, thank you

src/main/java/leo/task/TaskList.java
vigonometry: Updated to fit the Java coding standard (they should have the same level of indentation).

src/main/java/leo/task/Task.java
vigonometry: Updated method name to `setNotDone()`.

PR #254, opened 2023-01-25
[Lam Kai Yi Karen] iP

Karen-Lam:
# DukePro

> "Your mind is for having ideas, not holding them." - [David Allen](https://dansilvestre.com/productivity-quotes/)

### DukePro frees your mind of having to remember things you need to do. It's 
- text-based
- easy to learn
- ~~FAST~~  *SUPER*  FAST to use

### All you need to do is,
2. double-click it.
3. add your tasks.

### Features:
 - [x] Managing tasks
 - [ ] Reminders (coming soon)

##
```ruby
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -6,9 +6,14 @@ public Deadline(String description, String by) {
         this.by = by;
     }
 
+    public Deadline(String description, boolean done, String by) {
```
russelltankaimin: Perhaps consider renaming `done` to `isDone` as per the coding style. Boolean field names should start with a verb.

src/main/java/Deadline.java
```diff
@@ -6,9 +6,14 @@ public Deadline(String description, String by) {
...
+        super(description, done);
+        this.by = by;
+    }
+
     @Override
     public String toString() {
-        return "[D]" + super.toString() + " (by: " + by + ")";
+        return "D | " + super.toString() + " | by: " + by;
```
russelltankaimin: Perhaps consider using `String.format()` instead of just concatenating strings. It may be much concise. For your consideration 😊

src/main/java/Duke.java
```diff
@@ -1,11 +1,15 @@
...
 import java.sql.SQLOutput;
 import java.util.Scanner;
 import java.util.ArrayList;
+import java.io.File;
+import java.io.FileWriter;
+import java.util.regex.Pattern;
 
 public class Duke {
+    static String STR = "------------------------------------------------------------";
     static ArrayList<Task> arr = new ArrayList<Task>();
```
russelltankaimin: Perhaps a more fitting name instead of `arr` will be `tasks` or `taskLists` with emphasis on the plurality since it has a container for tasks.👍

src/main/java/Duke.java
```diff
@@ -15,19 +19,59 @@ public static void main(String[] args) {
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
 
+        try {
+            File f = new File("./data/duke.txt");
+            Scanner scannerOne = new Scanner(f);
+            while(scannerOne.hasNextLine()) {
+                String[] parts = scannerOne.nextLine().split(Pattern.quote(" | "));
+                switch (parts[0]) {
```
russelltankaimin: `switch` and `case` keywords should be indented on the same column

src/main/java/Duke.java
```diff
@@ -15,19 +19,59 @@ public static void main(String[] args) {
...
+                    case "T":
```
russelltankaimin: For tasks, maybe consider using Factory methods inside a parser. You can then encapsulate the parsing logic inside the parser. With that said, this may result in 
```java
case "D":
case "T":
case "E":
   parser.add(response);
   break;
```
Just a suggestion to keep code concise, but currently still looks good 👍👍. For your consideration.

src/main/java/AddDeadlineCommand.java
```diff
@@ -0,0 +1,56 @@
+package duke.command;
```
WesleyBLDC: the .java files all seem to be in one folder, recommend to put related files into its own folder for easier management and viewing.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+        return lst;
+    }
+
+    /**
+     * Converts the TaskList into a nicely formatted String with numbered bullet points.
+     *
+     * @return String The numbered List of Tasks.
+     */
+    public String toStringList() {
```
WesleyBLDC: Would this be redundant? With relation to the toString function that you already have.

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
...
+        return this.name.contains(word);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString() {
+        if (isDone) {
+            return "1" + " | " + this.name;
```
WesleyBLDC: It is not very clear what the 1 and 0 is supposed to represent. Perhaps there is a more intuitive way to show it or to update the javadocs.

WesleyBLDC (approved):
Overall the code and logic looks good.

The only thing is that the .java and classes are all in one folder. I think there could be better file management.

PR #256, opened 2023-01-25
[jeff] iP

jefrai:
# Project Miki 🎀✨

> The most merciful thing in the world, I think, is the inability of the human mind to correlate all its contents.
> <p align="right">-ef1500</p>

> Ogey rrat
> _<p align="right">P. Usada et al.</p>_

**Miki frees your mind from its physical locus.**
Miki is
- text-based
- [here](https://github.com/jefrai/ip/releases "there")
- transient
- eternal

**All you need to do is:**

1. download from [here](https://github.com/jefrai/ip/releases "miki/releases")!
2. `java -jar miki.jar`
3. add your tasks!
4. ???!
5. profit!

---

### With Project Miki, you can:

- [X] Plan your days with Miki!
- [ ] Autorun arbitrary files with Miki! (coming ~~soon~~ (time is subjective))
- [X] Get berated by Miki! (for poor spelling)
- [X] Procrastinate on your deadlines with Miki!
- [ ] save her

```java
    @Override
    public void run(TaskList tasks, Ui ui, Storage storage) {
        tasks.add(new Deadline("", LocalDateTime.of(2021, 05, 21, 15, 56, 25)));
        tasks.get(tasks.size() - 1).mark();
    }
```

> ashes to ashes, dust to dust
> _<p align="right">M. Ueda</p>_


src/main/java/Miki.java
```diff
@@ -0,0 +1,59 @@
+import command.Command;
+import shigure.Ui;
+import storage.Storage;
+import task.Parser;
+import task.TaskList;
+
+import java.io.IOException;
```
ExtraShotLatte: Great ordering of import statements!

src/main/java/Miki.java
```diff
@@ -0,0 +1,59 @@
...
+        Ui ui = new Ui(hasAsciiOnly);
+        Storage storage = new Storage("./data/");
+        TaskList tasks = new TaskList();
+        ui.printIntro();
+
+        if (!hasNoAutoload) {
+            try {
+                storage.load("autosave.txt", tasks);
+            } catch (IOException | Storage.MikiLoadException ex) {
+
```
ExtraShotLatte: Would it be better if the user can see that the loading has error?

src/main/java/Miki.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public static void main(String[] args) {
+        boolean hasAsciiOnly = false;
+        boolean hasNoAutoload = false;
+        for (int i = 0; i < args.length; i++) {
+            if (args[i].equals("--ascii-only")) {
+                hasAsciiOnly = true;
+            }
+            if (args[i].equals("--no-autoload")) {
+                hasNoAutoload = true;
```
ExtraShotLatte: I really like the feature of no autoloading of task!

src/main/java/command/Find.java
```diff
@@ -0,0 +1,29 @@
...
+        int matches = 0;
+        for (int i = 0; i < tasks.size(); i++) {
+            if (tasks.get(i).hasMatchingObjective(regex)) {
+                matches++;
+            }
+        }
+        ui.print("here's your " + matches + (matches == 1 ? " match:" : " matches:"));
+        for (int i = 0; i < tasks.size(); i++) {
+            if (tasks.get(i).hasMatchingObjective(regex)) {
+                ui.print(i + 1 + ". " + tasks.get(i));
```
ExtraShotLatte: Can we run the linear search only once?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,147 @@
...
+                    break;
+                case 'E':
+                    tasks.add(Event.parseLoad(repres));
+                    break;
+                default:
+                    //back-compat? try to handle
+                }
+            }
+        } catch (NumberFormatException | TaskParseException ex) {
+            throw new MikiLoadException("this file is corrupt...");
```
ExtraShotLatte: Corrupted instead of corrupt?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,139 @@
...
+     * Parses the supplied <code>String[]</code> command-line arguments to create a deadline.
+     *
+     * @param args arguments containing the deadline to be parsed.
+     * @return a deadline represented by <code>args</code>.
+     * @throws TaskParseException if <code>args</code> does not represent a valid deadline.
+     */
+    public static Deadline parseArgs(String[] args) throws TaskParseException {
+        String objective = "";
+        String by = "";
+        boolean isInTokenBy = false;
```
ExtraShotLatte: Perharps can explain what is the situation that this boolean is used for.

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,139 @@
...
+                + " " + (by.format(DATE_IN_FMT).codePoints().filter(c -> c == '\n').count() + 1);
+        repres.add(cur);
+        Collections.addAll(repres, objective.split("\n"));
+        Collections.addAll(repres, by.format(DATE_IN_FMT).split("\n"));
+        return repres.toArray(new String[repres.size()]);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
ExtraShotLatte, line 137: Great use of inheritDoc in here and other parts of the code!

ExtraShotLatte (commented):
Great personality 👍

LGTM! Great job in following the coding conventions as well!

src/main/java/Miki.java
jefrai: True - no message is raised if there's no autosave file since that's 'normal' when we're running for the first time,
but now if the file exists but is corrupt there's a new message to alert the user.

src/main/java/command/Find.java
jefrai: Yep, probably a good idea to keep the expensive regexes to a minimum - adjusted accordingly!

src/main/java/task/Deadline.java
jefrai: Added a little comment to that line & similar lines elsewhere 👍

src/main/java/storage/Storage.java
jefrai: 👀 Both are grammatical actually, but looks like corrupted is a lot more common when referring to files so switched to that

leonpoh (commented):
Looks good a lot of abstraction following single responsibility principle

PR #263, opened 2023-01-26
[dohaduong] iP

dohaduong:
### Duke 
**Your friendly task management app**

> "Until we can manage time, we can manage nothing else." - Peter Drucker [source](https://www.tameday.com/motivational-quotes-for-work-productivity/)

Your busy life deserves this.
**Duke** is 

- an _efficient_ task management app 💯 
- _easy_ to use 👍 
- helps you stay _organized_ 🥇 

To start using Duke, you only need to follow these **steps**: 

1. Download and install Duke
2. Add your tasks and deadlines
3. Let Duke handle your timetable for you!

**Available features:**

- [x] Managing tasks
- [ ] Managing deadlines and events
- [ ] Reminder for deadlines and events

Duke can be used to practice your Java as a programmer too! Here's the `main` method for you to start with: 
```ruby
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,199 @@
...
+    Bye,
+    List,
+    Mark,
+    Unmark,
+    Todo,
+    Deadline,
+    Event,
+    Delete,
+    Find
+}
```
pkpaing: Nice use of enum! Class/enum names are recommended to be nouns and written in PascalCase (e.g. Action instead of action)

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Returns the detail/content of deadline.
+     * Given that the detail is not empty.
+     *
+     * @param arr an array of the input.
+     * @return the String detail.
+     * @throws duke.MissingContentException if arr is empty.
+     */
+    public String deadlineDetail(String[] arr) throws MissingContentException {
```
pkpaing: Good use of camelCase for method names!

pkpaing (commented):
Looks good!

src/main/java/duke/command/Parser.java
```diff
@@ -0,0 +1,258 @@
...
+ * Makes sense of what users say
+ */
+public class Parser {
+
+    /**
+     * Makes sense of adding to do command from users
+     * @param arr array of original array
+     * @return full string command for users
+     */
+    public static String toDo(String[] arr) {
```
jugheadjones10: I think it might be more suitable if you name the method as a **verb** instead of a **noun** like "toDo". Maybe something like parseToDo? This way I will immediately know what the method does when I see the method name.

src/main/java/duke/command/TaskList.java
```diff
@@ -0,0 +1,242 @@
...
+        try {
+            try {
+                return this.arr;
+            } catch (IndexOutOfBoundsException e) {
+                throw new MissingContentException();
+            }
+        } catch (MissingContentException e) {
+            System.out.println(e.getMessage());
+        }
+        return this.arr;
```
jugheadjones10: I'm not entirely sure whether using nested try catch statements makes sense. An alternative would be using just one try catch statement, but in the catch statement you check for different types of errors and handle differently from there.

src/main/java/duke/command/TaskList.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Returns new task list.
+     * Unmarks task at given index as done.
+     *
+     * @param num1 index at which task need to be marked as undone.
+     * @return new task list with task unmarked.
+     * @throw InvalidIndexException if array at specific index is null
+     */
+    public String unmark(int num1) throws IOException {
```
jugheadjones10: Any particular reason you chose to use "num1" as the argument name? Seems a bit inconsistent with the other methods that simply use "num". Also, having a number in a variable name might be a bit unreadable.

jugheadjones10 (commented):
Great work! Some minor readability comments.

src/main/java/duke/command/TaskList.java
```diff
@@ -0,0 +1,250 @@
...
+                            + String.format("Now you have %d "
+                            + "tasks in the list", this.getValidLen()));
+                }
+            } catch (IndexOutOfBoundsException e) {
+                throw new InvalidIndexException();
+            }
+        } catch (InvalidIndexException e) {
+            return (e.getMessage());
+        }
+        return new InvalidIndexException().getMessage();
```
joytqt-1202: good catching of exception cases! 
but instead of nested `try-catch` blocks, could you try to use `try-catch-catch` instead?
`try (...) { ... } catch (...) { ... } catch (...) { ... }`

PR #266, opened 2023-01-26
[thedesalizes] iP

thedesalizes:


src/main/java/Duke.java
```diff
@@ -1,10 +1,101 @@
...
+                    System.out.println("bye ;;w;; see you again soon :3");
+                    System.out.println(line);
+                    running = false;
+                    break;
+                }
+                default: {
+                    System.out.println(line);
+                    System.out.println("bruh idk wtf that means");
+                    System.out.println(line);
+                }
```
kenzantonius: Note there is no indentation for case and default clauses.

src/main/java/Duke.java
```diff
@@ -1,10 +1,101 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+import tasks.*;
+import commands.*;
```
kenzantonius: Imported classes should always be listed explicitly.

kenzantonius (commented):
Overall, code is neat and follows most of the Java Coding Standard. Remember to make a Java Comments to make your code easier to read and understand :D

src/main/java/Duke.java
```diff
@@ -1,10 +1,101 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+import tasks.*;
```
elvern18: Understand that you need to import multiple classes from Task packages. However, import tasks.* has to be stated explicitly, such as import tasks.Task.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,38 @@
...
+
+    };
+
+    public Task(String description){
+        this.description = description;
+        this.isDone = false;
+    };
+
+    public String mark(){
+        isDone = !isDone;
```
elvern18: Will calling mark() on a task that is already marked cause the task to be unmarked?

elvern18 (commented):
LGTM, just need to tweak a few coding standards. Moreover, JavaDoc comments would be helpful in allowing other readers to understand your code better.

PR #272, opened 2023-01-26
[Ethan Chen Ee Shuen] iP

bipbipboopboop:
# Changee

> Click here to see the introduction of Changee

**Chagee** is **a simple to use task managing chat app**. You can tell **Chagee** to execute **commands** via **Command Line Interface** (CLI) or a **Graphical User Interface** (GUI).

Benefits of **Changee**:
- Create different types of tasks
- Show all tasks
- Delete tasks
- Mark as complete/incomplete
- Find tasks
- GUI

Roadmap:

- [X] Implement CLI
- [X] Implement GUI

You can also contribute to **Changee**. Locate `Laucher.java`, and start from this:

```java
public static void main(String[] args);
```


src/main/java/tasks/Tasks.java
```diff
@@ -0,0 +1,36 @@
+package tasks;
+
+public class Tasks {
```
Jnjy: Perhaps a more intuitive class name here?

Suggestion: `TaskList`

src/main/java/tasks/Tasks.java
```diff
@@ -0,0 +1,36 @@
...
+    private int numTasks = 0;
+    private Task[] tasksArray = new Task[100];
```
Jnjy: Perhaps you can use a more dynamic array here: `ArrayList<T>`

src/main/java/DukeParser.java
```diff
@@ -0,0 +1,116 @@
...
+
+        // Check whether the command is valid.
+        for (DukeCommand dukeCmd : DukeCommand.values()) {
+            if (dukeCmd.text.equals(command)) {
+                this.command = dukeCmd;
+                return;
+            }
+        }
+        // If command is invalid:
+        throw new Error("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
Jnjy: Any reason why you use the `Error` class instead of creating an `Exception`?

src/main/java/DukeCommand.java
```diff
@@ -0,0 +1,10 @@
+public enum DukeCommand {
+    BYE("bye"), MARK("mark"), UNMARK("unmark"), LIST("list"), TODO("todo"), DEADLINE(
+            "deadline"), EVENT("event");
```
Jnjy: I like how you extracted out possible commands and make it into an `Enum` class!

src/main/java/DukeParser.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private String[] parseCommandArgs() {
+        List<String> commandArgs = new ArrayList<>();
+        switch (command) {
+            case LIST:
+            case BYE:
+                break;
+
```
Jnjy: Any reason there is a breakline here?

Jnjy (commented):
I like how you adhere to Coding Standard in the early stage of IP. More extractions can be done to improve the OOP of the code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+
+    private static final String INDENTATION = "     ";
+    private static final String NEW_LINE = "\n";
+
+    private static void greet() {
+        System.out.println(BAR);
+        System.out.println(INDENTATION + "Hello! I'm Duke");
+        System.out.println(INDENTATION + "What can I do for you?");
+        System.out.println(BAR);
+        System.out.print(NEW_LINE);
```
winsonheng: Would it be more intuitive to use `System.out.println()` with no arguments instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+        System.out.println(BAR);
+        for (String text : texts) {
+            System.out.println(INDENTATION + text);
+        }
+        System.out.println(BAR);
+        System.out.print(NEW_LINE);
+    }
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
```
winsonheng, line 3: Should this be a final static field instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+                    }
+                    case UNMARK: {
+                        int taskIndex = Integer.parseInt(commandArgs[0]);
+                        echo("OK, I've marked this task as not done yet:",
+                                " " + tasks.unmarkTask(taskIndex));
+                        break;
+                    }
+                }
+                // System.out.println("Success! Your command is " + command.text);
+            } catch (Error e) {
```
winsonheng: Should you try to catch some other (and more specific) Exceptions instead? Maybe create a custom one? Errors generally indicate serious problems and it's probably [best not to attempt to catch them](https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html).

src/main/java/DukeParser.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+        // If command is invalid:
+        throw new Error("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+    }
+
+    private String[] parseCommandArgs() {
+        List<String> commandArgs = new ArrayList<>();
+        switch (command) {
+            case LIST:
```
winsonheng: Maybe consider adding a TODO comment to indicate that the code here is not ready?

PR #284, opened 2023-01-26
[zuohui48] ip

zuohui48:
# MrDuke

> "All models are wrong, but some are useful." -George Box

MrDuke helps you to micromanage all your tasks. It's
- easy to use 
- text-based

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE!**

You can: 
- [x] Manage tasks
- [x] Manage deadlines 
- [x] Set reminders

```java
public class Main {
    }
}
```


src/main/java/Parser.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+    public boolean checkList(String input) {
+        return input.split(" ")[0].equals("list");
+    }
+    public boolean checkBye(String input) {
+        return input.split(" ")[0].equals("bye");
+    }
+    public boolean checkDelete(String input) {
+        return input.split(" ")[0].equals("delete");
+    }
```
shaowi: Boolean variables/methods should be named to sound like booleans.

src/main/java/Storage.java
```diff
@@ -0,0 +1,105 @@
+import java.io.File;
+import java.io.IOException;
+import java.io.FileWriter;
+import java.io.*;
```
shaowi: Imported classes should always be listed explicitly.

src/main/java/Parser.java
```diff
@@ -0,0 +1,41 @@
+import java.util.Arrays;
+import java.util.stream.Stream;
+public class Parser {
+    private static String[] VALID_COMMANDS = {"mark", "unmark", "list", "delete", "bye", "todo", "event", "deadline"};
```
shaowi: final keyword can be specified for constants

src/main/java/TaskList.java
```diff
@@ -0,0 +1,136 @@
+import java.util.ArrayList;
+public class TaskList {
+    private ArrayList<Task> list;
```
shaowi: Possible to be more clear of what list is this ? :)

shaowi (commented):
Looks good to merge!

PR #300, opened 2023-01-26
[tingyic] iP

tingyic:


src/main/java/Initiator.java
```diff
@@ -0,0 +1,34 @@
...
+     * Throws DukeException if command is invalid.
+     *
+     * @param commands The commands from the user.
+     * @return appropriate commands.
+     * @throw DukeException if command is invalid.
+     */
+    public static Object InitiatorCommand(String commands) throws DukeException {
+        if (commands == "list") {
+            return new List();
+        } else if (commands == "mark") {
```
Oliverloo0909: nitpicking but will it adhere more to the coding standard if the argument name is "command" instead of "commands"?

src/main/java/Initiator.java
```diff
@@ -0,0 +1,34 @@
...
+public class Initiator {
+    /**
+     * Takes in a line of user input.
+     * If command is valid, return command accordingly.
+     * Throws DukeException if command is invalid.
+     *
+     * @param commands The commands from the user.
+     * @return appropriate commands.
+     * @throw DukeException if command is invalid.
+     */
```
Oliverloo0909: Interesting and neat way of handling cases :)

src/main/java/Task.java
```diff
@@ -0,0 +1,52 @@
+/**
+ * The Task class.
+ * Stores whatever tasks entered by the user and display them back to the user
+ * when requested.
+ * Acts as a superclass for all Task types.
+ */
+public class Task {
```
Oliverloo0909: Should this be abstract since there is never an instance of a task command?

src/main/java/Command.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Stores a list of possible commands a user can call on Duke.
+ * Acts as a superclass for all command types.
+ */
+public abstract class Command {
```
Oliverloo0909: I like how you used oop to encapsulate all your commands :)

src/main/java/Initiator.java
tingyic: agreed

src/main/java/Task.java
tingyic: agreed

src/main/java/Initiator.java
```diff
@@ -0,0 +1,34 @@
...
+        } else if (commands == "deadline") {
+            return new Deadline();
+        } else if (commands == "bye") {
+            return new Exit();
+        } else if (commands == "delete") {
+            return new Delete();
+        } else {
+            throw new DukeException("I am sorry but I failed to comprehend your command.");
+        }
+        return null;
```
kimberlybp: Perhaps there is no need for the `return null;` statement? Since your last else block will throw an exception, the code will not reach this last return statement. Its also best to avoid returning null sometimes to avoid NullPointerExceptions

src/main/java/Initiator.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Takes in a line of user input.
+     * If command is valid, return command accordingly.
+     * Throws DukeException if command is invalid.
+     *
+     * @param commands The commands from the user.
+     * @return appropriate commands.
+     * @throw DukeException if command is invalid.
+     */
+    public static Object InitiatorCommand(String commands) throws DukeException {
```
kimberlybp: Since this method only handles one command at a time, maybe `command` as the param name would be better?

kimberlybp (commented):
Overall, everything seems clean and your code complies with the coding standards, with some minor issues. lgtm ✌🏼

PR #303, opened 2023-01-26
[c0j0s] iP

c0j0s:
# Agent Duke  

- text-based
- easy to learn
- ~FAST~ _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://github.com/c0j0s/ip/releases).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)  
---
```java
public class Main {
    }
}
```


src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,64 @@
...
+
+import duke.utils.DateUtil;
+
+/**
+ * Event
+ */
+public class Event extends Task {
+
+    private LocalDateTime from;
+    private LocalDateTime to;
```
dohaduong: maybe can change this one to "startDate"/"endDate" so it's a little bit clearer? but it looks good anyways!

dohaduong (commented):
Looks good!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,119 @@
...
+    private LocalDateTime date2;
+
+    /**
+     * Constructor for adding a todo.
+     *
+     * @param tType  {@link DukeCommand} enum
+     * @param title  {@link String} object
+     * @param isDone boolean
+     */
+    public AddCommand(DukeCommand tType, String title, boolean isDone) {
```
pangrwa: The use of enums here to restrict the set of commands is fairly interesting!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Constructor for adding a eveny.
+     *
+     * @param tType  {@link DukeCommand} enum
+     * @param title  {@link String} object
+     * @param isDone boolean
+     * @param date1  {@link LocalDateTime} object
+     * @param date2  {@link LocalDateTime} object
+     */
+    public AddCommand(DukeCommand tType, String title, boolean isDone, LocalDateTime date1, LocalDateTime date2) {
```
pangrwa: Instead of writing one chunky block of code in an constructor, overloading constructors seems to be neater!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * AddCommand
+ */
+public class AddCommand extends Command {
+
+    private DukeCommand tType;
+    private String title;
+    private Boolean isDone;
+    private LocalDateTime date1;
```
pangrwa: Maybe perhaps you could have better variable naming for `date1` and `date2`. I'm not too sure which refers to the deadline or start/end date.

src/main/java/duke/constant/Message.java
```diff
@@ -0,0 +1,42 @@
+package duke.constant;
+
+/**
+ * Message strings of duke
+ */
+public class Message {
+    /* Exception messages */
```
pangrwa, line 7: I certainly like how you abstracted all the messages into one class. Makes cleaning the code much easier.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,168 @@
...
+                        DateUtil.toLocalDateTime(args[2]));
+            default:
+                throw new NoSuchCommandException(Message.EXCEPTION_NOSUCH_COMMAND);
+            }
+        } catch (DateTimeParseException e) {
+            throw new InvalidCommandArgsException(Message.EXCEPTION_INVALID_CMD_ARGS);
+        } catch (NumberFormatException e) {
+            throw new InvalidCommandArgsException(Message.EXCEPTION_INVALID_TASK_ID_FORMAT);
+        }
+
```
pangrwa, line 216: I like how you separate the code blocks based on the logic of arguments. However, since the method is rather long, you might want to consider separating the logic each of code block out into another methods?

src/main/java/duke/task/Event.java
pangrwa: I second this!

pangrwa (commented):
Overall, the logic of the code is great. I can see that you have abstracted as much as you could and I definitely learn a lot while reading your code. However, there are some areas you could improve on such as naming variables!

src/main/java/duke/task/Event.java
c0j0s: will certainly do.

PR #309, opened 2023-01-27
[abenx162] iP

abenx162:
# Page, the medieval chatbot!
- Are you looking for an app to keep track of your tasks?
- Do you want to have fun while doing it?

Use Page! Features:
1. add and delete Quests to your Quest Log
2. search Quest Log for keywords
3. and more!

Here's a behind-the-scenes look:
```Java
        Page page = new Page("data/questlog.txt");
        page.run();
    }
```

Keep track of tasks like this!
[ ] task 1
[x] task 2
[ ] task 3

Look at this review 👍 

> "I love using Page!" - Anonymous

Try it out [here](https://github.com/abenx162/ip/releases/tag/v0.1)!
hint: while using Page, type `help` to see a list of commands.

**THANK YOU** for reading!


src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Quest {
+    private LocalDateTime by;
```
wxxedu: Would it be better if we name this variable as `finishBy`?

src/main/java/Event.java
```diff
@@ -0,0 +1,21 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Quest {
+    private LocalDateTime from;
```
wxxedu: Can we change this to `fromDate` and `toDate` just so that the variable names could be more explicit?

src/main/java/Page.java
```diff
@@ -0,0 +1,222 @@
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+
+public class Page {
+
+    private Scanner scan;
```
wxxedu: Would it be better if we use `scanner` for the variable name? I've noticed similar issues in many other places as well.

wxxedu (commented):
I am wondering if you can add some doc comments as well. Also, I think it would be better to change some names in the code.

src/main/java/page/quest/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+
+import page.PageException;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Quest {
+    private LocalDateTime by;
+    private static DateTimeFormatter inputFormatter = DateTimeFormatter.ofPattern("[HHmm dd/MM/yy][dd MMM yyyy hh:mma]");
```
bofeng1999: This line may be abit too long, break it up into two lines if necessary

src/main/java/page/Storage.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Scanner;
+
+public class Storage {
+    private File storageFile;
+
+    public Storage(String filePath) {
+        this.storageFile = new File(filePath);
+    }
+
+    public ArrayList<Quest> loadData() throws PageException {
```
bofeng1999, line 37: Maybe add JavaDocs header comments, does not seems like a trivial method

src/main/java/page/Parser.java
```diff
@@ -0,0 +1,95 @@
+package page;
+
+import page.quest.*;
+
+public class Parser {
+
+    public boolean parseExecute(String input, Ui ui, Storage storage, QuestLog questLog) throws PageException {
```
bofeng1999: Add abstraction for this method, seems a bit too long

src/main/java/page/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+                    case "event":
+                        if (splitInput.length == 1) {
+                            throw new PageException("Sorry, please include a task description!");
+                        } else {
+                            String[] splitFrom = splitInput[1].split(" /from ");
+                            if (splitFrom.length == 1) {
+                                throw new PageException("Sorry, please include a /from time!");
+                            } else {
+                                String[] splitTo = splitFrom[1].split(" /to ");
+                                if (splitTo.length == 1) {
```
bofeng1999: Very deep nesting here, a bit hard to read, maybe can extract out certain code as a different method

bofeng1999 (commented):
LGTM! Just some minor improvements needed! :)

PR #310, opened 2023-01-27
[StefanosAv]iP

StefanosAv:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
```
Yufannnn: I like how the indentation and placement of braces in this class follow good programming practices, making the code easier to read and understand.  The code is well-structured and easy to read. The class name, variables and methods are named clearly and appropriately. :wink:

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.lang.Math;
+
 public class Duke {
```
Yufannnn: I like how you have used comments in the code to explain the logic and make it easier to understand. However, any reason why you structured the method this way instead of breaking it down into simpler methods or keeping the methods shorter and more readable? :worried:

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+                }
+            } else if (command.startsWith("mark ") || command.startsWith("unmark ")) {
+                try {
+                    // Mark task as done
+                    if (command.substring(0, command.indexOf(" ")).equalsIgnoreCase("mark")) {
+                        int index = Integer.parseInt(command.replaceAll("mark ", "")) - 1;
+                        if (taskList.get(index).isDone()) {
+                            throw new DukeException("marked");
+                        }
+                        else {
```
Yufannnn: should the line ```` else { ```` be placed on the same line with the preceding `}`, according to the code standard for if-else statement? :frowning:

src/main/java/DukeException.java
```diff
@@ -0,0 +1,44 @@
+public class DukeException extends Exception {
```
Yufannnn: Any reason why you did it this way instead of using polymorphism or having separate classes for each exception?

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
+public class Task {
```
Yufannnn: Any reason why you did not make the Task class abstract since you did not directly instantiate a Task object anyway?

Yufannnn (commented):
Overall, the code is mostly easy to follow, but there are a few places with overly nested logic. Additionally, there are a few minor deviations from common coding standards. It is great to see that the code is well-commented on, making it easier to understand. Some of your methods are getting lengthy, making them difficult to read and maintain. Consider breaking them down into smaller, simpler methods.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public void WrongCommandException() {
+        if (this.task.equals("wrong")) {
+            System.out.println("OOPS!!! I'm sorry, but I don't know what that means.");
+        }
+    }
+
+    public void EmptyDescriptionException() {
+        switch (this.task) {
```
unfazing: You can consider amending the indentation rules for the `case` blocks according to the Java conventions (no indent in front of `case`)

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+                    }
+                    // Create new Deadline task
+                    else if (command.toLowerCase().startsWith("deadline ")) {
+                        String description = command.substring(command.indexOf(" ") + 1, command.indexOf("/"));
+                        String by = command.substring(command.indexOf("/by") + 4);
+
+                        System.out.println(description + " - " + by);
+
+                        if (description.equals("")) {
+                            throw new DukeException("deadline");
```
unfazing: You may want to include a more specific error message when you throw Exceptions.

unfazing (commented):
The code is generally readable and follows the coding standard!

PR #337, opened 2023-01-28
[John Chia Heng Kit] iP

Johnchiahk:
# Duke project template

> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes/)

Duke frees your mind of having to remember things you need to do. It's 

- text-based
- easy to learn

All you need to do is, 

1. download it from [here](https://nus-cs2103-ay2223s2.github.io/website/schedule/week4/project.html)
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [X] Managing deadlines
- [ ] Reminders (coming soon)
---
```java
public class Main {
    }
}
```


wengkit1 (commented):
Could exceptions be possibly extracted and handled in another package?  
Otherwise nice design lgtm!

sembcorpp:
Code is easy to read and follow but maybe some java docs would be good so its more clear what the method is doing and what params it takes in? But good job so far

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,192 @@
...
+                    for (int i = 0; i < taskStorage.noTasks(); i++) {
+                        System.out.print(i + 1 + ".");
+                        System.out.println(taskStorage.getTask(i));
+                    }
+                    break;
+
+                case "bye":
+                    System.out.println("Byeee! Hope to see you again! Signing off, duke.Duke.");
+                    break;
+
```
kimberlybp: Do remember that coding standards state that here should be no indentation for `case` clauses in switch statements

src/main/java/duke/TaskStorage.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public TaskStorage() {
+        taskstorage = new ArrayList<Task>();
+    }
+    public void addTask(Task t) {
+        System.out.println("Got it. I've added this task:");
+        System.out.println("  " + t);
+        this.taskstorage.add(t);
+        ind++;
+        System.out.println("Now you have " + ind + " task(s) in the list.");
```
kimberlybp: Perhaps you can move System.out.println(..) into a function in your Ui class to utilise it more, since the purpose of the Ui class is to handle all interactions with the user

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,192 @@
...
+//        String logo = " ____        _        \n"
+//                + "|  _ \\ _   _| | _____ \n"
+//                + "| | | | | | | |/ / _ \\\n"
+//                + "| |_| | |_| |   <  __/\n"
+//                + "|____/ \\__,_|_|\\_\\___|\n";
+        String line = "-------------------------------";
+//        System.out.println(line);
+//        System.out.println("Hiii Im\n" + logo);
+//        System.out.println("What can I do for you hmm?");
+//        System.out.println(line);
```
kimberlybp: A small suggestion: would recommend removing commented, unused lines of code (for cleaner code)

kimberlybp (commented):
Overall, code looks good with some small issues

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,192 @@
...
+                        taskStorage.deleteTask(eventTask);
+                    } catch (NumberFormatException e) {
+                        System.out.println("Number should be typed in");
+                    } catch (DukeException e){
+                        System.out.println(e.getMessage());
+                    } finally {
+                        break;
+                    }
+
+                default:
```
wengkit1: good to include default behaviour, consistent with coding standards.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,192 @@
...
+                        if (taskNo > taskStorage.noTasks() || taskNo <= 0) {
+                            throw new DukeException("Give a vaild number");
+                        }
+                        Task eventTask = taskStorage.getTask(taskNo - 1);
+                        taskStorage.deleteTask(eventTask);
+                    } catch (NumberFormatException e) {
+                        System.out.println("Number should be typed in");
+                    } catch (DukeException e){
+                        System.out.println(e.getMessage());
+                    } finally {
```
wengkit1: Is a finally block necessary in this case?

src/main/java/duke/Duke.java
Johnchiahk: Thanks I've rectified it

src/main/java/duke/Duke.java
Johnchiahk: Alright thank you!

src/main/java/duke/TaskStorage.java
Johnchiahk: Noted with thanks!

src/main/java/duke/Duke.java
Johnchiahk: Hi wengkit! On second thought, the finally block is not necessary, but I believe the break statement clearly shows to the reader that each case will only be executed at most once 😅

src/main/java/duke/Duke.java
Johnchiahk: Thank you!

PR #338, opened 2023-01-29
[rachtan27] iP

rachtan27:
# Duke
> I'm not a lazy person, I just get things done when I want to. - [Unknown](https://tfiglobalnews.com/2022/06/04/lazy-quotes/)

Sounds like you? Don't worry because Duke is here to help you get your life in order by remembering the things you need to do. It':
- text-based
- easy to learn 
- ~~fast~~ **_SUPERFAST_** to use

**All you need to do is:**
1. download it
2. add your tasks
3. let duke do the rest!  👍 

**Features**
- [x]  Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)  

If you are a Java programmer, you can use it to practice Java too. Here's the main method:
```ruby
public class Duke {
        new Duke("data/duke.txt").run();
    }

### CODE REVIEW

File: src/main/java/Duke.java [hunk 1]
```diff
@@ -1,10 +1,189 @@
+import java.util.Scanner;
+import java.util.ArrayList;
 public class Duke {
-    public static void main(String[] args) {
+    public static void main(String[] args) throws DukeException {
```
hongshenggg: The main method is quite long imo. Would it be possible to extract some code into other classes/methods?

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
+import java.util.Scanner;
+import java.util.ArrayList;
 public class Duke {
```
hongshenggg: Good job on following the naming conventions of variables in this class

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task{
```
hongshenggg: Good job on following the naming conventions

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
+//addapted from partial solution provided in lecture notes
```
hongshenggg: Misspelled adapted here.

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
...
+//https://nus-cs2103-ay2223s2.github.io/website/admin/ip-w2.html
+public class Task {
+    String desc;
+    String type;
+    boolean isDone;
+
+    public Task(String desc) {
+        this.desc = desc;
+        this.isDone = false;
+        this.type = null;
```
hongshenggg: Is there a better way than initializing this variable to null?

hongshenggg (commented):
LGTM. Just some nits to fix.

PR #353, opened 2023-01-29
[minosx31] iP

minosx31:
# DukePro
> “Your mind is for having ideas, not holding them.” – David Allen [(Source)](https://dansilvestre.com/productivity-quotes/)

- text-based
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is __FREE!__

Features:

- [X] Managing tasks
- [ ] Reminders (coming soon)

***

```java
public class Main {
  }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+
+        @Override
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    private static String numOfTasks(ArrayList<Task> tasks) {
+        return tasks.size() == 1 ? " task " : " tasks ";
+    }
```
LiuMC-SG: The method name could perhaps be changed as it is misleading. numOfTasks would normally imply returning the number of tasks the list contain. However, you are just getting the string of whether it task should be plural or not. Method names should also start with a verb instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+        Scanner sc = new Scanner(System.in );
+
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello I'm Duke!\n" + "How can I help you?");
```
LiuMC-SG: Maybe you can take out the string in your println to avoid "magic" strings. Such as welcomeMsg = "...". There are several instances later on as well that you can take not of.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public Deadline (String description, String deadline) {
+        super(description);
+        this.doBy = deadline;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + "[" + super.getStatusIcon() + "] " + super.getDescription() + " (by: " + this.doBy + ")";
+    }
```
LiuMC-SG: I feel like there is no need to use `super.getStatusIcon` and `super.getDescription` since your Deadline already extends from Tasks. Hence, these methods already exist within the Deadline class. Furthermore, the use of this should be removed if there is no multiple instances of it. This is a suggestion.
```suggestion
    public String toString() {
        return String.format("[D] [%s] %s (by: %s)", getStatusIcon(), getDescription(), doBy);
    }
```
The other classes that extend from Task have similar problem and can be dealt with respectively as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+                } else if (command.equals(commands.UNMARK.toString())) {
+                    int indexOfTask = Integer.parseInt(remainingInput) - 1;
+                    Task currentTask = tasks.get(indexOfTask);
+                    currentTask.markAsUndone();
+                    System.out.println("OK, I've marked this task as not done yet:");
+                    System.out.println("[" + currentTask.getStatusIcon() + "] " + currentTask.getDescription());
+                } else if (command.equals(commands.TODO.toString())) {
+                    if (remainingInput == null) {
+                        throw new IncorrectInputException("Enter a description!");
+                    }
```
LiuMC-SG: I like that you did this so that it is easier to see the "happy path" that the code would traverse through.

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+                if (command.equals(commands.BYE.toString())) {
+                    System.out.println("Bye. Hope to see you again!");
+                    break;
+                } else if (command.equals(commands.LIST.toString())) {
+                    System.out.println("Here are the tasks in your list");
+                    if (tasks.size() != 0) {
+                        for (int i = 0; i < tasks.size(); i++) {
+                            System.out.println((i + 1) + "." + tasks.get(i).toString());
+                        }
+                    }
```
LiuMC-SG: Perhaps the if condition can be taken out so that it is easier to see the "happy path" and reduce indentation of code.
```suggestion
                    if (tasks.size() == 0) {
                        continue;
                    }
                    for (int i = 0; i < tasks.size(); i++) {
                        System.out.println((i + 1) + "." + tasks.get(i).toString());
                    }
```

LiuMC-SG (commented):
Overall okay so far, just a few key points to take note of to change. Some of the changes such as simplifying the long if-else/switch case would come naturally after the next few levels.
Keep up the good work! :D

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+    public enum commands {
+        BYE, LIST, MARK, UNMARK, TODO, DEADLINE, EVENT, DELETE;
+
+        @Override
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    private static String numOfTasks(ArrayList<Task> tasks) {
```
venuslimm: I agree with @LiuMC-SG, perhaps this can be a suggestion?
```suggestion
    private static String getTaskString(ArrayList<Task> tasks) {
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+        return tasks.size() == 1 ? " task " : " tasks ";
+    }
+
     public static void main(String[] args) {
+        ArrayList<Task> tasks = new ArrayList<>();
+        Scanner sc = new Scanner(System.in );
```
venuslimm: Should there be no extra space after `System.in`?
```suggestion
        Scanner sc = new Scanner(System.in);
```

src/main/java/Event.java
```diff
@@ -0,0 +1,13 @@
...
+    private String[] period;
+
+    public Event(String description, String[] period) {
+        super(description);
+        this.period = period;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + "[" + super.getStatusIcon() + "] " + super.getDescription() + " (from: " + this.period[0] + " to: " + this.period[1] + ")";
```
venuslimm: Would it be better to use line break as the character count exceeds 120?

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+                    String deadline = deadlineDetails.length > 0 ? deadlineDetails[1] : null;
+
+                    if (deadline == null) {
+                        throw new IncorrectInputException("Enter a valid deadline!");
+                    }
+                    Task newTask = new Deadline(deadlineDescription, deadline);
+                    tasks.add(newTask);
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println(newTask);
+                    System.out.println("Now you have " + tasks.size() + Duke.numOfTasks(tasks) + "in the list.");
```
venuslimm: I like how the different logical units are separated by a line! The same goes for the rest of this if statement.

venuslimm (commented):
LGTM! I have commented on a few suggestions but generally, I think your code is easy to read! I like how you extracted the Task class very neatly into the 3 other classes.

PR #355, opened 2023-01-29
[benjamin-wee] iP

benjamin-wee:
DukeMaster helps you keep track of your tasks, so you can finally hit your deadlines and reach your goals. It is:
+ intuitive 
+ command-line friendly
+ ~~very~~ **ULTRA** simple to use 

You're *this* close to getting your life together!

You simply have to:

1. Download DukeMaster [here](https://github.com/benjamin-wee/ip/releases/tag/A-Jar)
2. Start up DukeMaster
3. Log your tasks
4. And that's it!

You won't have to worry about forgetting tasks anymore! 🤩 

Features:

- [ ] Manage deadlines
- [ ] Manage events
- [ ] Manage ToDos

If you are a Java Programmer, it's easier than printing "Hello World" with the `main` method shown below!
```Java
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,215 @@
+import java.util.*;
```
wongyx: Perhaps it would be better to list out the specific imports that you need from java.util

src/main/java/Duke.java
```diff
@@ -1,10 +1,215 @@
...
+                        + list.size() + " tasks in the list.");
+            } else {
+                try {
+                    if (input.contains("todo")) {
+                        String description = input.replace("todo", "");
+                        Todo todo = new Todo(description);
+                        list.add(todo);
+                        System.out.println("Got it. I've added this task:\n" + todo.toString() + "\nNow you have "
+                                + list.size() + " tasks in the list.");
+                    } else if (input.contains("deadline")) {
```
wongyx: Not sure if my logic is wrong but would it be better to check if input starts with deadline instead of contain? If somehow the input is "deadline make a todo app /by 2023/02/01" does it cause the code to excercute the previous if statement block of todo?

```suggestion
                    } else if (input.startsWith("deadline")) {

**Review by wongyx — COMMENTED** [review]
There might be some minor issues with the code but overall it is great!!

File: src/main/java/Storage.java [hunk 3]
```diff
@@ -0,0 +1,98 @@
+
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.io.File;
+import java.io.FileNotFoundException;
```
junlee1991: The ordering of import statements must be consistent. (alphabetically)

src/main/java/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+        } else if (type.equals("deadline")) {
+            try {
+                String[] descriptionDate = taskDescription.split("/by ");
+                String description = descriptionDate[0];
+                String date = descriptionDate[1];
+                LocalDate dateString = LocalDate.parse(date);
+                Deadline deadline = new Deadline(description, dateString);
+                taskList.add(deadline);
+                System.out.println("Got it. I've added this task:\n" + deadline.toString() + "\nNow you have "
```
junlee1991: Should put this in the UI since it has to do with printing

src/main/java/Duke.java
```diff
@@ -1,10 +1,49 @@
+import java.util.Scanner;
+
 public class Duke {
```
junlee1991: Overall, I found your code easy to read and understand!

junlee1991 (commented):
Good Job! keep it up!

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,40 @@
+package duke;
+
+public class Task {
+    protected String description;
+    protected Boolean isDone;
+
+    public Task(String description) throws MissingDescriptionException {
+        if (description.equals("") || description.equals(" ")) {
```
jednhk, line 20: Do you need to consider the case where the description is two spaces or more?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,125 @@
...
+        this.taskList = new ArrayList<Task>();
+    }
+
+    public void addTask(String taskDescription, String type) {
+
+        if (type.equals("todo")) {
+            try {
+                Todo todo = new Todo(taskDescription);
+                taskList.add(todo);
+                System.out.println("Got it. I've added this task:\n" + todo.toString() + "\nNow you have "
```
jednhk: Perhaps move this to the UI class? Since it has to do with user interaction

jednhk (commented):
LGTM. Perhaps will be easier to debug in the future if the print statements are in the Ui class instead.

src/main/java/duke/MissingDescriptionException.java
```diff
@@ -0,0 +1,13 @@
+package duke;
```
bryanongjx, line 1: good idea creating a missingdescriptionexception class!

src/main/java/duke/Task.java
bryanongjx, line 20: can consider using isEmptyString instead !

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,40 @@
+package duke;
+
+public class Task {
```
bryanongjx, line 8: can consider making this class abstract! not sure though, but worth a thought!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public String toStorageData() {
+        String completed = this.getStatusIcon();
+        DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("d MMM u");
+        String dateString = date.format(dateFormatter);
+        return taskType + "//" + completed + "//" + description + "//" + dateString;
+    }
+}
```
bryanongjx: can consider adding documentation for your classes

PR #356, opened 2023-01-30
[jeraldkiew] iP

jeraldkiew:


src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+            String str = sc.nextLine();
+            if (str.equals("bye")) {
+                System.out.println("Leo: Good talk man, catch you again some other time!");
+                break;
+            } else if (str.equals("list")) {
+                System.out.println("Leo: Here you go!");
+                for (int i = 0; i < tasks.size(); i++) {
+                    System.out.println(i + 1 + ". " + tasks.get(i));
+                }
+            } else if (str.split(" ")[0].equals("mark")) {
```
nerdyboy98: I think for this "mark" block, you can do a try/catch block to catch IndexOutOfBoundException in case the user input is just "mark" instead of "mark 1". Applicable to other blocks that require a subsequent input after the keywords like unmark, delete, etc. as well :D

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
...
+        this.isDone = false;
+        return this.toString();
+    }
+
+    /**
+     * Custom toString includes checkbox showing whether task is done
+     * @return custom toString
+     */
+    @Override
+    public String toString() {
```
nerdyboy98: I love how the printing of "X" and " " is implemented here by utilising the isDone boolean as a logic check.

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
+import java.util.*;
```
nerdyboy98: Wildcard imports should be avoided so that readers at a glance can tell what are the stuff that you have imported.

nerdyboy98 (commented):
Overall its rather clean and there's definitely a couple of takeaways for me as well.

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public String unmark() {
+        this.isDone = false;
+        return this.toString();
+    }
+
+    /**
+     * Custom toString includes checkbox showing whether task is done
+     * @return custom toString
+     */
```
zhoyx: Do note that a line break should be added between the method summary and return field.

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
+public class Task {
+    private boolean isDone;
```
zhoyx: I like that your boolean naming follows the best practices 👍

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
...
+    private final String str;
```
zhoyx: Maybe a more descriptive name such as "description" could improve clarity?

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+                System.out.println("Leo: New deadline added:");
+                System.out.println(tasks.get(tasks.size()-1));
+                System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+            } else if (str.split(" ")[0].equals("event")) {
+                String[] spaceSplit = str.split(" ", 2);
+                String[] fromSplit = spaceSplit[1].split("/from", 2);
+                String[] toSplit = fromSplit[1].split("/to", 2);
+                tasks.add(new Event(fromSplit[0], toSplit[0], toSplit[1]));
+                System.out.println("Leo: New event added:");
+                System.out.println(tasks.get(tasks.size()-1));
```
zhoyx: Adding a space (`tasks.size() - 1`) might improve readability.

zhoyx (commented):
I only noticed some minor issues but other than those, overall the code looks great 👍

PR #360, opened 2023-01-30
[Jeremykhoo] iP

MrJeremyKhoo:
# DukeExtended
duke now can:
`remember to do things - duke `
- store tasks
- remember deadline
- ~~help you be productive~~
all you need is :
1. [time
](https://en.wikipedia.org/wiki/Time)2. patients to **learn**
there are even task lists:
- [ ] like this
- [x]  and this :shipit:

main is 
```java
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+
+	protected String by;
+	public Deadline(String description, String by) {
+		super(description);
+		this.by = by;
+	}
+
+	@Override
+	public String toString() {
+		return "[D]" + super.toString() + " (by: " + by + ")";
```
Junyi00: I believe this file has extra level of indentations here weirdly. I notice the same issue occurring in other files too...

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+	private static void markTask(int i, boolean b) {
+		int index = i - 1;
+		list.get(index).markTask(b);
+		System.out.println("Marked/Unmarked the task, task is in the state:");
+		System.out.print("  " + list.get(index));
+	}	
+	
+	private static void addList(Task task) {
+		list.add(task);	
+		System.out.println("added: " + task.getDescription());
```
Junyi00: Could it better if we just use the task's string representation here instead? This would consequently remove the need for an extra getter in the Task's class

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+		System.out.print("You have: " + list.size() + " task(s)");
+	}	
+					
+	private static void deleteTask(int i) {
+		int index = i - 1;
+		System.out.println("removed: " + list.get(index).toString());
+		System.out.print("You have: " + (list.size() - 1)+ " task(s)");
```
Junyi00: Missing trailing whitespace before '+'. This is a common issue found in other part of the codebase too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+						deadline = String.join(" ", l);
+						addList(new Deadline(description,deadline));
+					}	
+					catch (IllegalArgumentException e) {
+						throw new DukeException(e);	
+					}
+
+					catch (Exception e) {
+						throw new DukeException(e);	
+					}
```
Junyi00: How about shifting your `try-catch` logic wrapped around the switch-case block instead? That would reduce some code duplication when the same exception could be thrown

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+		}
+		
+		line(in.length());
+		String[] parmArr = in.split("\\s+");
+		List<String> parm = Arrays.asList(parmArr);
+		try {
+			parseIn(parm);
+		}
+		catch (DukeException e) {
+
```
Junyi00: I see that the DukeException will send a message to user when it is created, I believe adding a comment for this behaviour would be helpful for others to understand why there is an empty catch block.

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+		System.out.print("You have: " + (list.size() - 1)+ " task(s)");
+		list.remove(index);	
+
+	}
+	
+	private static void parseIn(List<String> parm) throws DukeException {
+			int byIndex;
+			int fromIndex;
+			String description;
+			List<String> l;
```
Junyi00: `List<String> l` could be given a more descriptive variable name, similar for variable names like `f` and `t`

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
+import java.util.Scanner; 
+import java.util.ArrayList; 
+import java.util.List; 
+import java.util.Arrays; 
 public class Duke {
-    public static void main(String[] args) {
+	private static ArrayList<Task> list;
```
Junyi00: This is an interesting concept of ensuring only one instance of task list exists. Any reason it is designed this way? If  new Duke was somehow initialised, the array list is 'resetted' base on how the constructor is written now

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
+public class Task {
+	protected String description;
+	protected boolean isDone;
+
+	public Task(String description) {
+		this.description = description;
+		this.isDone = false;
+	}
+
+	public String getStatusIcon() {
```
Junyi00: This could be set to private since it is only used internally within the class

Junyi00 (commented):
Great Effort!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
+
+	protected String by;
```
kohkaixun: There seems to be extra indentation for every line in not just this file, but the other files as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,211 @@
...
+					catch (IllegalArgumentException e) {
+						throw new DukeException(e);	
+					}
+
+					catch (Exception e) {
+						throw new DukeException(e);	
+					}
+				break;
+			case "event":
+					try{
```
kohkaixun: Maybe leave a space between the try and starting curly bracket?

kohkaixun (commented):
There are probably a few nits to take care of before merging. But those nits are mostly coding standards stuff like more specific names, indentation and spacing.

PR #363, opened 2023-01-30
[ryanchua00] iP

ryanchua00:
Roody Application
---
> "Your mind is for having ideas, not holding them." - David Allen

The Roody application helps frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPERFAST to use!

All you need to do is,

1. download it from [here](https://github.com/ryanchua00/ip/releases)
2. double click it
3. add your tasks from the **command line**
4. let it manage your tasks for you :yum:

And it is **FREE**!

Features:

- [X]  Managing Tasks
- [X]  Managing Events
- [ ]  Reminders! (Coming soon)

If you Java programmer, you can use it to practice Java too. Here's the `constructor` for Roody method:

```
/**
  * Creates a chatbot with specified filepath to task data.
  * @param filepath The filepath to task data.
  */
public Roody(String filepath) {
    // Assumed no more than 100 tasks
    this.list = new ArrayList<Task>();
    this.ui = new Ui();
    this.storage = new Storage(filepath);
}
```


src/main/java/Roody.java
```diff
@@ -152,13 +163,25 @@ public static void main(String[] args){
             // else, repeat
             } else if (inputs[0].equals("list")) {
                 roody.printList();
+            } else if (inputs[0].equals("delete")) {
+                if (inputs.length == 2) {                    
+                    roody.delete(inputs[1]);
+                } else {
+                    new RoodyException("Please enter a index number to be deleted");
```
peaol: Maybe can try using switch statements instead of if-statements here to have a neater code 😄 
the method split() can be helpful in determining "delete", "list" or "mark" cases etc!

src/main/java/Roody.java
```diff
@@ -3,15 +3,13 @@
 import java.util.Scanner;
```
peaol: Nice, you followed the coding standards for import statements!

peaol (commented):
Good job on implementing Roody.java, perhaps adding some more OOP like classes would be good!

src/main/java/Roody.java
ryanchua00: Thanks for this, made it smoother.

src/main/java/roody/Roody.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public Roody(String filepath){
+        // Assumed no more than 100 tasks
+        this.list = new ArrayList<Task>();
+        this.ui = new Ui();
+        this.storage = new Storage(filepath);
+    }
+    
+    // toggles completion status of tasks
+    private void complete(int index, boolean complete) throws RoodyException{
```
lennoxtr: perhaps a javadoc @param comment here would be helpful to explain why boolean complete is needed

src/main/java/roody/Roody.java
```diff
@@ -0,0 +1,130 @@
...
+                task.setDone();
+            } else {
+                task.setUnDone();
+            }
+            ui.showMarkStatus(complete, task);
+        }
+    }
+    
+
+    private void delete(int index) {
```
lennoxtr: A more intuitive name here would be great
```suggestion
    private void deleteTask(int index) {
```

lennoxtr (commented):
Your code is intuitive and very easy to read. Good Job!

PR #364, opened 2023-01-30
[JThh] ip

JThh:
### DukeBro

> “Try harder, fail harder, and win harder.” – Myself ([source](https://myownquotes/#1))

What you need to be successful is simply a task manager - helps you take down all your deadlines and not miss any of them.

- Feature 1: text-based
- Feature 2: low learning curve
- Feature 3: fast to pick up and use for programmers

All you need to do is,

1. download it from [here](https://github.com/JThh/ip/archive/refs/tags/release-v0.2.zip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x]  Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
...
+            super(msg);
+        }
+    }
+
+
+    public static String parse_date(String s) {
+        DateTimeFormatter read_fmt = DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+        DateTimeFormatter print_fmt = DateTimeFormatter.ofPattern("MMM dd yyyy");
+        try {
+            LocalDate lt = LocalDate.parse(s, read_fmt);
```
NappySprout: Is it possible to make the argument more expressive?

src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
...
+        protected String by;
+
+        public Deadline(String description, String by) {
+            super(description);
+            this.by = by;
+        }
+
+        @Override
+        public String toString() {
+            return "[D]" + super.toString() + " (by: " + by + ")";
```
NappySprout: To keep it simple you can use string formatting to prevent accumulation of '+' in string concatenation
```java
int i = 461012;
System.out.format("The value of i is: %d%n", i);
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
...
+
+
 public class Duke {
+    public static ArrayList<Task> todos = new ArrayList<>();
+    public static File dir = new File("./data/");
+    public static File file = new File("./data/todo_list.txt");
+
+    public static class Task {
+        protected String description;
+        protected boolean isDone;
```
NappySprout: good job at consistent naming of the booleans!

src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
...
+import java.time.format.DateTimeParseException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+import java.io.File;
+
+
 public class Duke {
+    public static ArrayList<Task> todos = new ArrayList<>();
```
NappySprout: Good naming of arrays and making them plural!

src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+import java.io.File;
+
+
```
NappySprout: remember to separate the files and provide javadoc when you reach the appropriate stages 😄

NappySprout (commented):
The code is neatly written and the switch statement is methodic.

The classes present like Task has been abstracted out so splitting them into different files will not be too troublesome and will be easy for you.

src/main/java/Duke.java
```diff
@@ -96,6 +96,7 @@ public DukeEmptyTaskException (String msg) {
     public static String parse_date(String s) {
```
YZmunchmunch: Instead of "parse_date", perhaps you could rename it to "parseDate", which conforms to the coding standard?

PR #366, opened 2023-01-30
[Truong Minh Duong] iP

Zhongli5712:
# Duke
> It's not a bug; it's an undocumented feature. [source](https://techvify-software.com/35-best-coding-programming-quotes/).
## Allows you to

* Add new task
* Mark/Unmark a task is done
* Delete a task
* Find a task

## Type of tasks
1. Todo task 
2. Deadline task
3. Event task

## Special features :cowboy_hat_face: :
- [x] **Auto** save and load previous task list
- [x] Print user-friendly syntax error
- [ ] Motion avatar (comming soon)

### For developer, can run the `main` method in Duke class:
```java
        new Duke().run();
    }
```


WindyCall (commented):
Generally, not enough efforts put into CS2103T and not enough codes for review.
I will do the back up one instead.

src/main/java/duke/TodoList.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public void mark(int index) throws DukeExceptions{
+        int todo_list_length = todo_list.size();
+        if (index < 0 || index > todo_list_length) {
+            throw new DukeExceptions("Please use list command to check the index!");
+        }
+        Task task = todo_list.get(index - 1);
+        task.markTask();
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println(todo_list.get(index-1));
```
kimberlybp: Perhaps you can move `System.out.println(..)` into a function in your Ui class to utilise it more, since the purpose of the Ui class is to handle all interactions with the user

src/main/java/duke/DukeExceptions.java
```diff
@@ -0,0 +1,12 @@
+package duke;
+
+public class DukeExceptions extends Exception{
```
kimberlybp, line 6: While the class name is valid according to coding standards, perhaps it would be more intuitive if it was named `DukeException` instead since it only represents an instance of 1 Exception

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,17 @@
...
+    private TodoList todoList;
+    private int index;
+
+    public MarkCommand(TodoList todoList, int index) {
+        this.todoList = todoList;
+        this.index = index;
+    }
+
+    @Override
+    public void execute() throws DukeExceptions { todoList.mark(index); }
```
kimberlybp: Just a small suggestion✌🏼 Since the MarkCommand and UnmarkCommand classes do the same thing except the execute functions maybe you want to consider just using 1 class that handles all marking related commands with a check to determine whether to mark or unmark the task

src/main/java/duke/TodoList.java
```diff
@@ -0,0 +1,67 @@
...
+        int todo_list_length = todo_list.size();
+        if (index < 0 || index > todo_list_length) {
+            throw new DukeExceptions("Please use list command to check the index!");
+        }
+        Task task = todo_list.remove(index - 1);
+        System.out.println("Noted. I've removed this task:");
+        System.out.println(task);
+    }
+
+    public int number_of_tasks() {
```
kimberlybp: Coding standards state the names representing methods must be written in camelCase, so i think `getNumberOfTasks` or `numberOfTasks` would be better

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,187 @@
...
+
+        public Event(String task) throws DukeExceptions {
+            super();
+            String[] commands = task.split(" /from ");
+            if (commands.length == 1) {
+                throw new DukeExceptions("OOPS!!! Looks like someone forget when the event begins :)\n Please use /from to indicate begin time");
+            }
+            this.task = commands[0];
+            String[] from_to_timeline = commands[1].split(" /to ");
+            if (from_to_timeline.length == 1) {
```
kimberlybp, line 168: I've noticed some of your variable names use snake case, however the coding standards states that variable names must be in camelCase, so do remember to check and update your variable names

kimberlybp (commented):
Other than some minor coding standard issues, the code generally looks neat with good logic 👍🏼

PR #368, opened 2023-01-31
[Lin Weilin] iP

weilin1202:
# BearyBear
>“Your mind is for having ideas, not holding them.” – David Allen [source](https://dansilvestre.com/productivity-quotes/)

BearyBear frees your mind of having to remember things you need to do. It's,

- text-based
- **easy** to learn
- *SUPER FAST* to use

## All you need to do is,

1. *download* it from [here](https://github.com/weilin1202/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

## Features:
- [x] Manage Different Types of Tasks
- [x] Update status of tasks
- [x] Save the tasks locally 

---------------------------------------------------------------------------------------------------------------------
```ruby
public class Launcher {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
+
+    String taskDescription;
+    LocalDate deadLine;
```
nicleejy: Since these class fields have corresponding getters and setters, it might be better to make them private to prevent any unintended modifications.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+                LocalDate deadline = LocalDate.parse(input.substring(input.indexOf("/") + 4));
+                Deadline task = new Deadline(input, deadline);
+                taskList.add(task);
+                addTaskToFile("C:/Users/linwe/Documents/TaskSaved.txt", task.getTaskType() + " | "
+                        + task.currentTaskStatus()  + " | " + task.getTask() + " | " + task.getDeadline() + "\n");
+
+                System.out.println("Got it. I've added this task: \n  " + task +
+                        "\nNow you have " + taskList.size() + " tasks in the list.");
+
+            } else if(input.length() > 5 && (input.substring(0, 5)).equals("event")) {
```
nicleejy: Try to surround reserved words with a whitespace for consistency

src/main/java/Event.java
```diff
@@ -0,0 +1,30 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task{
+    String taskDescription;
+    LocalDate startDate;
+    LocalDate endDate;
```
nicleejy: Same goes for this, regarding making class fields private

src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public String getDeadline() {
+        return this.deadLine.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " ( by: " + this.getDeadline() + " )" ;
+    }
```
nicleejy: This is a good way to separate the different display behaviors of different task types 👍

nicleejy (commented):
Overall, the coding standard has been followed, just apart from some minor issues! 👍 In most of your classes, do remember to keep fields private to prevent unwanted modifications by the user. It might also be good to declare Task as an abstract class since it should not be instantiated at any point in the program. Do let me know if you have any queries regarding my comments :)

PR #370, opened 2023-01-31
Increments for iP

Shuggan:
# Duke
> "My laptop before me, and the world at my fingertips. Take pride coder" - Coder's creed [source](https://drive.google.com/file/d/1fsPKB-ANbXKy7eP5qOAudEwgTO2t4PNA/view?usp=sharing)
### Duke frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- FAST SUPER FAST to use
- Full of cute profile images

All you need to do is,
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😉 hehe!

And it is FREE! Wow!

Features:

- [x] Managing tasks
 - [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+            if (textInput.length() >= 4 &&
+                    textInput.substring(0, 4).equalsIgnoreCase("todo")) {
+                try {
+                    String[] parts = textInput.split(" ", 2);
+                    if (parts.length == 1 || parts[1] == "") {
+                        throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+                    }
+                    Task t = new Task.Todo(textInput.substring(5));
+                    taskList.add(t);
+                    String output = String.format("Got it. I've added this task:\n%s\nNow you have %d tasks in the list", t.toString(), taskList.size());
```
joellow88: This output line seems very long, is it possible to split them into 2 lines?
A similar issue appears in line 32, 78, 88.

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+            if (textInput.equalsIgnoreCase("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
+                scan.close();
+                return;
+            }
+
+            if (textInput.equalsIgnoreCase("list")) {
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 0; i < taskList.size(); i++) {
+                    System.out.println(i + 1 + ". " + taskList.get(i).toString());
```
joellow88: This manual string formatting works, however, have you considered using Java's builtin String.format method?

src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+public class DukeException extends Exception{
```
joellow88: For this class and your Task class, can you indent it with 4 spaces instead of 2?

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+            }
+
+            if (textInput.length() >= 8 &&
+                    textInput.substring(0, 6).equalsIgnoreCase("delete")) {
+                int i = Integer.parseInt(textInput.substring(7));
+                Task t = taskList.get(i - 1);
+                taskList.remove(i - 1);
+                String output = String.format("Got it. I've removed this task:\n%s\nNow you have %d tasks in the list", t.toString(), taskList.size());
+                System.out.println(output);
+                continue;
```
joellow88: I find the use of multiple `continue` instructions in the code strange. The while loop already has a looping behaviour once it reaches the end of the loop, so using `continue` to return to the start of the loop should not be a default behaviour. What is your rationale for designing it this way?

Rather, I would suggest the following structure:
```java
boolean isRunning = true;
while (isRunning) {
    if (textInput.equalsIgnoreCase("bye")) {
        isRunning = false;
    } else if (stmt) {
        code;
    } else {
        code;
    }
}
```

Alternatively, you may consider parsing the command first, then using a switch-case statement to handle various cases.

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
+import java.util.ArrayList;
```
joellow88: Would prefer if the code was in a package, eg: package duke.

joellow88 (commented):
Up to Week 2 (Level-6), the code is functional and meets all the weekly requirements. There are minor issues regarding the coding standard, and a few suggestions for refactoring in the Duke.java class.

PR #376, opened 2023-01-31
[Qiu Qianhui] iP

QQH0828:
# Duke

> “Your mind is for having ideas, not holding them.” – David Allen ([Source](https://dansilvestre.com/productivity-quotes))

Duke is a chatbot that frees your mind from having to remember the things you need to do. It's

- Text-Based
- Easy to learn
- ~~Fast~~ _Super Fast_ to use

All you need to do is, 👍 

1. Download it from [here](https://github.com/QQH0828/ip/releases/tag/v0.1).
2. Double-click it
3. Add your tasks
4. Let it manage your tasks for you 

And it is ****FREE****! :partying_face:

Features: 👍 

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders
- [ ] GUI interface

---

If you are a random programmer that's looking through this, you can use it to practice Java as well. Here is the `main` code for the launcher.

``` js
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -45,22 +67,51 @@ public void addTask(Task task) {
...
+        }
+
+        command = command.split(" ", 2)[1];
+        if (command.startsWith("/by")) {
+            throw new DukeException("☹ OOPS!!! The description of a deadline cannot be empty.\n");
+        } else if (command.endsWith("/by")) {
+            throw new DukeException(("☹ OOPS!!! The due date of a deadline cannot be empty.\n"));
+        }
+
+        String info[] = command.split(" /by ");
```
ShanHng: Perhaps we should put the [] right after the String instead even though this is valid too?

src/main/java/Duke.java
```diff
@@ -45,22 +67,51 @@ public void addTask(Task task) {
...
         String description = info[0];
         String by = info[1];
```
ShanHng: Maybe instead of just "by", we can use a variable name that indicates that this is a date?

src/main/java/Duke.java
```diff
@@ -45,22 +67,51 @@ public void addTask(Task task) {
...
+            throw new DukeException("\t ☹ OOPS!!! The from date or due date of a deadline cannot be empty.\n"
+                    + "\t Or the format is wrong.\n");
+        }
+
+        command = command.split(" ", 2)[1];
+        if (command.startsWith("/from") || command.endsWith("/to") || command.contains("/from /to")) {
+            throw new DukeException("\t ☹ OOPS!!! The description, from date or due date of a event cannot be empty.\n");
+        }
+
+        String info[] = command.split(" /from ");
```
ShanHng: Perhaps the naming of this variable can be a standard plural noun instead?

src/main/java/Duke.java
```diff
@@ -14,43 +14,88 @@ public void start() {
...
+    public void deadline(String command) {
+        String info[] = command.split(" ", 2)[1].split(" /by ");
+        String description = info[0];
+        String by = info[1];
+        Deadline deadline = new Deadline(description, by);
+        addTask(deadline);
+    }
+
+    public void event(String command) {
+        String info[] = command.split(" ", 2)[1].split(" /from ");
```
ShanHng: Ooh I like your clever usage of split to distinguish between /to argument and /from argument!

src/main/java/Duke.java
```diff
@@ -14,43 +14,88 @@ public void start() {
...
+        for (int i = 0; i < taskIndex; i++) {
+            System.out.print("\t " + (i + 1) + "." + tasks[i].toString() + "\n");
         }
+        System.out.print("\n");
+    }
+
+    public void addTask(Task task) {
+        tasks[taskIndex] = task;
+        taskIndex++;
+        System.out.println("\t Got it. I've added this task:\n"
```
ShanHng: I like what you did here - indenting and spacing a sentence into parts that are more readable!

src/main/java/Duke.java
```diff
@@ -14,43 +14,88 @@ public void start() {
...
+                + "\n\t Now you have " + taskIndex + " tasks in the list.\n");
+    }
+    public void todo(String command) {
+        String description = command.split(" ", 2)[1];
+        Todo todo = new Todo(description);
+        addTask(todo);
+    }
+
+    public void deadline(String command) {
+        String info[] = command.split(" ", 2)[1].split(" /by ");
```
ShanHng: Perhaps we could use a variable name that indicates this is an array? Also, remember to have the square brackets right after the data type~! (String[])

ShanHng (commented):
Code is clean and easy to read! But there is repeated violation of the rule to have the [ ] placed behind the data type instead of variable name. Generally looks great!

src/main/java/Duke.java
QQH0828: Thanks for pointing them out!

PR #377, opened 2023-02-01
[Fahim Tajwar] iP

fahim-tazz:
# DukePro
> \"Your mind is for having ideas, not storing them.\" - [David Allen](https://dansilvestre.com/productivity-quotes)

DukePro frees your mind off remembering tasks. It's  
- Text-based
- easy to learn

And it's **free**!

All you need to do is:
1. Download Duke from [here](https://github.com/fahim-tazz/ip/releases/tag/A-JAR).
2. Go to Terminal and enter the following:
```sh
cd ~/Downloads/         #or wherever your .jar file is.
java -jar duke.java
# Enjoy!
```
> Double-clicking the .jar file can corrupt it on MacOS, since MacOS flags it as unsafe.

3. Add your tasks with commands : `todo`, `deadline` and`event`.
4. Mark tasks as done or undone with `mark` and `unmark`.
5. Delete unwanted tasks with `delete`.
6. Can't find the task you're looking for? Just use `find`!
7. And let Duke handle the rest! :bowtie:

Features:
- [x] Managing todo-list
- [x] Managing deadlines
- [x] Tracking future events
- [ ] Reminders (coming soon)


ajjajjajjajj (commented):
Good work here, some suggestions for your code:
- Wildcard imports shouldn't be used, this might've resulted from the default settings in Intellij that auto converts imports into wildcard imports (try disabling it)
- You might want to implement an `symbol` attribute to each type of task, rather than hardcoding it into the `toString()` method. This would also help for the storage requirement, if you want to use the task symbols in the storage format :)
- A `reply(String reply)` method in your Duke class can solve the tediousness of having to print the divider line for each output by the chatbot
- Each case for a switch statement should be on the same indent line as the `switch():` statement itself, there's also a setting in Intellij that disables that

Nice job though, I found your code easy to read and understand :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,122 @@
...
+    //        System.out.println("Hello from\n" + logo);
+
+    String divider = "____________________________________________________________\n";
+    System.out.println(divider + "Hello! I'm Duke");
+    System.out.println("What can I do for you?\n" + divider);
+
+    ArrayList<Task> tasks = new ArrayList<>();
+    Scanner sc = new Scanner(System.in);
+
+    while(sc.hasNext()) {
```
jedidiahC: I noticed several Java reserved words such as `while` in this case not being followed by a white space. e.g.
```suggestion
    while (sc.hasNext()) {
```
Perhaps you could consider using IntelliJ's code cleanup feature to add these white spaces automatically.

src/main/java/Duke.java
```diff
@@ -1,10 +1,122 @@
...
+      String instr = sc.nextLine();
+      switch(instr) {                                 //Single-word instructions
+        case "list":  //List down
```
jedidiahC: Should indentation for `case` clauses be removed?
```suggestion
    case "list":
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,122 @@
...
+
 public class Duke {
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+  public static void main(String[] args) {
```
jedidiahC: Should basic indentation be changed from 2 spaces to 4 spaces to align with the Java coding standard?

```suggestion
    public static void main(String[] args) {

**Review by jedidiahC — COMMENTED** [review]
Overall, the code seems okay. However, you may want to consider some of Java's coding standards in regards to layout and comments.

File: src/main/java/Duke.java (hunk 1 above)
Comment (arkarsg) [inline src/main/java/Duke.java]: Be careful of using `sc.hasNext()` vs `sc.hasNextLine()`

File: src/main/java/Duke.java [hunk 4]
```diff
@@ -1,10 +1,122 @@
...
+          String keyword = "";  //keyword = argument(s) after the action.
+          while(tokens.hasMoreTokens()) {
+            keyword = keyword + " " + tokens.nextToken();
+          }
+          keyword = keyword.strip();
+          if(keyword.equals("")) {
+            System.out.println(divider + "OOPS!!! The description of a " + action + " cannot be empty.\n" + divider);
+          }
+
+          switch(action) {                            //For instructions with argument(s).
```
arkarsg: Would recommend against nesting `switch-case` block for code quality

src/main/java/Duke.java
```diff
@@ -1,10 +1,122 @@
...
+                System.out.println("Sorry, the index number you've entered does not exist.");
+                continue;
+              }
+              break;
+            default:
+              System.out.println(divider + "OOPS! I'm sorry, but I don't know what that means :-(\n" + divider);  //For unknown action.
+          }
+      }
     }
+  }
```
arkarsg: A lot of nested blocks. Consider abstracting away with methods within the main class. For example you can have:
```java
// in main
switch(...) {
    case "todo":
        addToDo(tokens);
}

// add implementation
public void addToDo(String[] tokens) {
// your implementation
}
```

arkarsg (commented):
Good start so far. Consider abstracting away with more functions as the code expands with future levels.

PR #399, opened 2023-02-06
Use assert to document important assumptions

zxisatree:
Use assert statements to verify invariants and check that the application is working correctly at runtime.


zxisatree:
Wrong PR to upstream repo.

